PR #8, opened 2020-05-25
Make @webauthntine/browser work in node env

### LINKED ISSUE #6: @webauthn/browser not working with SSR
`@webauthn/browser` needs access to `window` when loading the module.
This causes errors in SSR environments, because window is not defined there.
Would be great if loading the module would also work in SSR environment.

P4sca1:
This is useful when `@webauthntine/browser` is used with SSR, e.g. using `nuxt.js` or `next.js`.

The first problem was that `window` is undefined in node which causes the `supportsWebauthn` method to throw an error trying to index undefined. This has been fixed by using typescript optional chaining.

The second problem is that the umd build is not working in node. This issue was discussed in https://github.com/webpack/webpack/issues/6784 and the solution is to use `output.globalObject: this` as also stated in [the docs](https://webpack.js.org/configuration/output/#outputglobalobject).

Closes #6

### DIFF
--- a/packages/browser/src/helpers/supportsWebauthn.ts
+++ b/packages/browser/src/helpers/supportsWebauthn.ts
@@ -3,7 +3,7 @@
  */
 export default function supportsWebauthn(): boolean {
   return (
-    window.PublicKeyCredential !== undefined
+    window?.PublicKeyCredential !== undefined
     && typeof window.PublicKeyCredential === 'function'
   );
 }
--- a/packages/browser/src/helpers/supportsWebauthn.test.ts
+++ b/packages/browser/src/helpers/supportsWebauthn.test.ts
@@ -1,4 +1,4 @@
-import supportsWebauthn from './supportsWebauthn';
+import supportsWebauthn from './supportsWebauthn'
 
 beforeEach(() => {
   // @ts-ignore 2741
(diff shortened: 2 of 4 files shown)


P4sca1:
@MasterKale

packages/browser/src/helpers/supportsWebauthn.ts
```diff
@@ -3,7 +3,7 @@
  */
 export default function supportsWebauthn(): boolean {
   return (
-    window.PublicKeyCredential !== undefined
+    window?.PublicKeyCredential !== undefined
```
MasterKale, line 6: Can you add a unit test to **supportsWebauthn.test.ts** for this please?

P4sca1: I added a test.

P4sca1:
Btw would be great to use something like prettier or eslint styling rules to enforce a common coding style.

MasterKale:
And yes, you're absolutely right about this project needing some code standards. I plan on adding at least Prettier formatting via `husky` and `lint-staged`, I just wasn't expecting collaborators so soon 🤗

P4sca1:
I messed up the last commit. Could you remove the `if (global.window)` check in `setupTests.ts` again?

MasterKale:
@P4sca1 Done and done, I ran the test suite and nothing adverse happened after removing it :v:

Also this fix has been published as **@webauthntine/browser@0.2.2**.

PR #9, opened 2020-05-25
Add support for excludeCredentials

P4sca1:
Can be used to prevent a single webauthn authenticator to be used multiple times for a single user account, which may impose a security risk (https://github.com/w3c/webauthn/issues/1235#issuecomment-501646115).

https://www.w3.org/TR/webauthn/#dictdef-publickeycredentialdescriptor

https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/WebAuthn_Client_Registration.html

### DIFF
--- a/packages/server/src/helpers/convertCOSEtoPKCS.test.ts
+++ b/packages/server/src/helpers/convertCOSEtoPKCS.test.ts
@@ -1,8 +1,6 @@
-import cbor from 'cbor';
 import { COSEKEYS } from '@webauthntine/typescript-types';
-
+import cbor from 'cbor';
 import convertCOSEtoPKCS from './convertCOSEtoPKCS';
-
 
 test('should throw an error curve if, somehow, curve coordinate x is missing', () => {
   const mockCOSEKey = new Map<number, number | Buffer>();
--- a/packages/browser/src/helpers/toPublicKeyCredentialDescriptor.ts
+++ b/packages/browser/src/helpers/toPublicKeyCredentialDescriptor.ts
@@ -1,0 +1,22 @@
+import base64js from 'base64-js';
+
+/**
+ * A helper method to convert a base64 encoded credential id into a PublicKeyCredentialDescriptor,
+ * which is usable by allowCredentials and excludeCredentials.
+ */
+export default function toPublicKeyCredentialDescriptor(
+    base64CredentialId: string,
+    transports?: AuthenticatorTransport[]
+): PublicKeyCredentialDescriptor {
+    // Pad id with = characters until it is a multiple of 4 so it is a proper base64 encoded string.
+    // This is required because credential.id returned by navigor.credentials.get/create is base64 encoded but not padded.
+    const padLength = 4 - base64CredentialId.length % 4;
+    base64CredentialId = base64CredentialId.padEnd(base64CredentialId.length + padLength, '=');
+
+    return {
+      type: 'public-key',
+      id: base64js.toByteArray(base64CredentialId),
+      transports
+    };
+}
+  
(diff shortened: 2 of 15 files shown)


P4sca1:
This is still WIP.

P4sca1:
@MasterKale I am having problems implementing this.
I think I found the root cause for my issues. Converting an `ArrayBuffer` to base64 and then converting back to `UInt8Array` leads to a result where the initial value and the decoded value differ from each other.

Take the following example:
```ts
  const credential = await navigator.credentials.create({ publicKey }) as PublicKeyCredential;

  if (!credential) {
    throw new Error('Attestation was not completed');
  }

  const rawId = credential.rawId
  const encodedRawId = toBase64String(rawId)
  const decodedRawId = toUint8Array(encodedRawId)
  console.log(rawId, encodedRawId, decodedRawId, rawId == decodedRawId)
```

The code above leads to the result that rawId does not equal decodedRawId.
I don't know if I can compare `ArrayBuffer` and `UInt8Array` in such a way, but I basically need  a way to get the rawId, encode it to store it in database and later convert it back to the same `ArrayBuffer` so that `navigator.credentials.get` finds it.

MasterKale:
[`credential.id` is already the base64-encoded version of `credential.rawId`](https://w3c.github.io/webauthn/#publickeycredential) - in fact it's what I pass back from browser's `startAssertion()` to server's `verifyAssertionResponse()`.

Using `id` would remove your first two lines since it's already, literally, "encoded raw ID". Can you use that instead? And since it's been sufficient to pass `credential.id` through `base64js.toByteArray()` to create the "PublicKeyCredentialDescriptor" for `allowCredentials` in an assertion (see **browser**'s `startAssertion()`), I think it'll be sufficient for an attestation's `excludeCredentials` because [it's the same type in the spec](https://w3c.github.io/webauthn/#dictdef-publickeycredentialdescriptor).

P4sca1:
Id is base64url encoded. I thought that would cause incompatibility with the base64 helper. However your code in startAssertion indeed does exactly what I need to do. I will try it out tomorrow. Thanks for your help.

MasterKale:
Take note, I found you need to pad the base64 string you pass into `base64js.toByteArray()` with `=` characters because it expects proper base64 strings, with a length that's a multiple of four. I noticed that `credential.id` _is_ base64, but is _not_ padded, but adding `=` till it is works just fine.

See how I craft `allowCredentials` in **browser's** `startAssertion()`. This'd probably need to be moved into a new "toPublicKeyCredentialDescriptor(base64CredentialID)" helper if it works for attestations' `excludeCredentials` too.

P4sca1:
Is the `authenticatorInfo.base64CredentialID` returned by `verifyAttestationResponse` (which originates from `credential.response.attestationObject`) the same as `authenticatorInfo.base64CredentialID` returned by `verifyAssertionResponse` (which originates from `credential.id`)?

P4sca1:
I created a new `toPublicKeyCredentialDescriptor` helper.
`excludeCredentials` functionality is done and it works on my local test page.
For some reason `allowCredentials` does not work on my local test page. It always says that no authenticator is registered for this site. I still need to investigate whether that is an issue with my implementation or with the library.
During development I noticed that it might be good to support the `userVerification` and `rpId` settings in `generateAssertionOptions` and `generateAttestationOptions`.

packages/server/src/helpers/convertCOSEtoPKCS.test.ts
```diff
@@ -1,9 +1,7 @@
-import cbor from 'cbor';
 import { COSEKEYS } from '@webauthntine/typescript-types';
```
MasterKale, line 1: This should probably become an `import type` as well.

P4sca1:
Implemented in #11

PR #11, opened 2020-05-26
feature/exclude-credentials-attestation

MasterKale:
This PR updates `generateAttestationOptions()` with the ability to pass in a list of credential IDs to exclude from the attestation process (thanks to @P4sca1 for getting the ball rolling on this). If an authenticator has already been used to perform an attestation, the authenticator will simply throw an error in the browser when it is asked to perform another attestation and an internal ID matches an entry in the attestation option's `excludeCredentials`:

Screen Shot 2020-05-25 at 11 56 43 PM

`generateAttestationOptions()` and `generateAssertionOptions()` also gained the ability for suggested authenticator transports to be passed in. In both cases these will default to suggesting all types of authenticators.

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -104,6 +104,7 @@ app.get('/generate-attestation-options', (req, res) => {
      * The username can be a human-readable name, email, etc... as it is intended only for display.
      */
     username,
+    devices,
   } = user;
 
   /**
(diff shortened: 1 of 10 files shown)


P4sca1:
Great work! Now only rpId and userVerification are missing. Chrome will always complain when userVerification is not explicitly set.

Issue #20, opened 2020-06-01
Simplify @simplewebauthn/browser


MasterKale:
Having been exposed to some other existing libraries, I've thought of ways to improve the **browser** package:

**1. Return _everything_ from `navigator.credentials.create()` and `navigator.credentials.get()`**

Right now the library picks and chooses what to send back to the back end. This'll likely not scale if/when WebAuthn expands, especially as more and more extensions get added.

It also means **server** is tightly-coupled to **browser** (an intentional decision, admittedly) - having **browser** return the entire response from the `navigator.credentials` commands (and updating **server** to accept the full response) will offer a more decoupled pair of packages in case a better alternative to **browser** exists/comes along while **server** still remains a more superior implementation of the back end.

Exposing more info to the **server** package might also be needed for more spec-compliant attestation/assertion verification - right now verification seems pretty reliable, but it's not a to-the-letter implementation.

I feel this'll also make conformance testing easier as it'll mean less proxy code needs to be written for the various tests they run code through.

**2. Try to remove the single project dependency**

This'd basically involve cribbing some notes from Github's https://github.com/github/webauthn-json. It turns out that if we lean more on modern browsers' `Uint8Array` support and the global `atob` method, we can completely do away with third-party dependencies. I'd like to continue to make **browser** as lean as possible, and it looks like we can make it even slimmer than it already is.

MasterKale:
These changes were addressed in #21. These improvements are now available in v0.5.0 of the libraries.

PR #21, opened 2020-06-02
Refactor Megamix 1

MasterKale:
This is a pretty extensive PR intended to make it easier to pass info from **browser** to **server** by exposing more of the data structures returned from the **browser** start methods to the complimentary verification methods in **server**. This should make it easier to track what values are being returned from these calls and enhance them with future return values as the spec evolves. It should also make it easier to test these libraries for conformance with the spec.

For more context on the original intent of this PR, see #20.  

There was also some mixed use of "base64" encoding when, [according to the spec](https://w3c.github.io/webauthn/#sctn-dependencies), "base64url" encoding is required. This was due to me, in my naivety, conflating the two. It should now be easier to tell when these encoded values are being used versus the ArrayBuffers they represent.

**For more detailed information on the changes in this PR, see below:**

- **[browser]** (BREAKING) Refactor `startAttestation()` and `startAssertion()` to return more of the output from the `navigator.credentials` calls
- **[browser]** Replace `base64-js` dependency with internal functionality
- **[browser, server]** Standardize on use of Base64URL encoding when converting to and from JSON
- **[server]** (BREAKING) Remove references to "base64" from `generateAttestationOptions()` and `generateAssertionOptions()` by renaming the `excludedBase64CredentialIDs` and `allowedBase64CredentialIDs` to `excludedCredentialIDs` and `allowedCredentialIDs` respectively
- **[typescript-types]** (BREAKING) Migrate some non-shared typings into **server**


MasterKale:
These changes have been released as v0.5.0

Issue #39, opened 2020-07-28
Error prone toUint8Array method


mahnunchik:
I've faced that `toUint8Array` method treats input as ASCII string: https://github.com/MasterKale/SimpleWebAuthn/blob/bc9ad0f68fc49c4ac23cd22428248faa26d3f9b6/packages/browser/src/helpers/toUint8Array.ts#L6

It is impossible to change transformation of `challenge` on the client.

1. Attestation https://github.com/MasterKale/SimpleWebAuthn/blob/bc9ad0f68fc49c4ac23cd22428248faa26d3f9b6/packages/browser/src/methods/startAttestation.ts#L27
2. Assertion https://github.com/MasterKale/SimpleWebAuthn/blob/bc9ad0f68fc49c4ac23cd22428248faa26d3f9b6/packages/browser/src/methods/startAssertion.ts#L27

Whats wrong?

First: Challenge verification will fail if challenge string contains non ASCII character. For example: `abcж` yep, `ж` occupy 2 bytes.

```js
const value = 'abcж';

const array = Uint8Array.from(value, c => c.charCodeAt(0));
const string = String.fromCharCode.apply(null, array);

/// 'abcж' !== 'abc6'
value !== string
```

Second: user may want to transfer challenge from server in different format:

* ASCII string - current assumption
* HEX string - I prefer this format
* OCTET string
* base64 or base64url strings
* and so on

So it would be better to allow user configure challenge encoding or move conversion and appropriate helper(s) to userland instead hard coded into the lib.

MasterKale:
Thank you for reporting this, I think it's a fair point that the library should support UTF-8 strings. I'll look to fix that in the next release.

Regarding other potential challenge formats: the primary reason ASCII is assumed is because challenges need to be transmittable as `application/json` content. While I'm willing to concede the point that UTF-8 encoding can be considered de facto and so should be supported, I hesitate to commit to introducing support for arbitrary encoding/decoding of challenges. When a simple ASCII guid is sufficient as a challenge, where's the value in the library supporting myriad other encodings? The name of the library is **Simple**WebAuthn, after all 😄

MasterKale:
UTF-8 values in challenges should now be supported in the just-released **@simplewebauthn/browser** v0.7.3.

mahnunchik:
Hi, thank you.

[...]

PR #40, opened 2020-07-29
bugfix/utf8-challenge

MasterKale:
This PR updates **@simplewebauthn/browser** to add support for the use of UTF-8 values in challenges from the server.


MasterKale:
This should resolve #39.

PR #42, opened 2020-07-30
feature/better-challenges

MasterKale:
This PR incorporates feedback in #39 to improve challenges support:

- **[server]** The `challenge` parameter of `generateAttestationOptions()` and `generateAssertionOptions()` is now _optional_.
  - **When undefined** the library will generate a random challenge. This value will be base64url-encoded in preparation for transit to the front end.
  - **When defined** the value will be directly encoded to base64url in preparation for transit to the front end.
- **[server]** `verifyAttestationResponse()` and `verifyAssertionResponse()` now require the above base64url-encoded challenge to be passed in as `expectedChallenge`.
- **[browser]** `startAttestation()` and `startAssertion()` now convert the base64url-encoded `options.challenge` to a buffer before passing it to the authenticator.

### DIFF
--- a/packages/browser/src/methods/startAssertion.test.ts
+++ b/packages/browser/src/methods/startAssertion.test.ts
@@ -4,6 +4,8 @@ import {
 } from '@simplewebauthn/typescript-types';
 
 import supportsWebauthn from '../helpers/supportsWebauthn';
+import toUint8Array from '../helpers/toUint8Array';
+import bufferToBase64URLString from '../helpers/bufferToBase64URLString';
 
 import startAssertion from './startAssertion';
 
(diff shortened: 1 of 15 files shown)


mahnunchik:
Awesome! It solves the issue #39 and keeps the library simple.

Just a small performance improvement:

> The crypto.randomBytes() method will not complete until there is sufficient entropy available. This should normally never take longer than a few milliseconds. The only time when generating the random bytes may conceivably block for a longer period of time is right after boot, when the whole system is still low on entropy.
> https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback

As a `randomBytes` may take a time, then it is suitable to not generate the `challenge` if is provided in the options.

Some kind of:

```js
if (!options.challenge) {
  options.challenge = generateChallenge();
}
```

mahnunchik:
Yep, you are right!

Issue #44, opened 2020-08-03
Improve findability


mahnunchik:
It should make it easier to find the library:

* Add NPM keywords https://docs.npmjs.com/files/package.json#keywords
* Add Github topics https://github.blog/2017-01-31-introducing-topics/

At least `FIDO` and `WebAuthn` may be added 😉

MasterKale:
NPM keywords are already taken care of:

<img width="503" alt="Screen Shot 2020-08-03 at 7 24 23 AM" src="https://user-images.githubusercontent.com/5166470/89193758-50a5ba00-d55b-11ea-9578-54bb0fc19852.png">

Now that the library is conformant I'll `FIDO` to these too. And **browser** may need a couple more too, now that I'm reviewing what it currently has.

As for Github topics, good call, I'll add some to the repo.

MasterKale:
The repo now has some appropriate topics:

<img width="322" alt="Screen Shot 2020-08-03 at 8 19 58 AM" src="https://user-images.githubusercontent.com/5166470/89198502-299eb680-d562-11ea-9fdf-9a4b1e56f253.png">

I'll get `fido` added to the three packages in the next release, whenever that happens.

mahnunchik:
Awesome 👍

MasterKale:
The latest deployment of v0.8.2 included updated **npm** listing info with the new `fido` tag:

<img width="487" alt="Screen Shot 2020-08-11 at 10 44 36 PM" src="https://user-images.githubusercontent.com/5166470/89979316-4c952e80-dc24-11ea-84c4-361b3ed5bc40.png">

Issue #47, opened 2020-08-07
requireResidentKey Android 7 bug


mahnunchik:
Hi,

I was trying to catch the bug for a week: Android 7.0 requires `authenticatorSelection.requireResidentKey` to be set to `false` to works without error:

> An unknown error occurred while talking to credential manager

Yep, it is default value https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialCreationOptions/authenticatorSelection but attestation doesn't work without it.

![photo_2020-08-07_12-47-26](https://user-images.githubusercontent.com/780935/89638477-2e3dd480-d8ac-11ea-87a1-07bc49037b82.jpg)

I think the library may set `authenticatorSelection.requireResidentKey` to default value `false`.

MasterKale:
Huh, so Android 7 won't support attestations unless that optional value is explicitly set in attestation options? I don't see an issue with explicitly initializing that value in the library, might as well aim for as much device support as possible.

Can I know what device and specific OS variant you're testing on (stock Android, etc...)? I don't have anything that old to test with anymore so I can't reproduce this myself but I'm still curious.

MasterKale:
@mahnunchik I've got PR #48 out that should fix this. Are you comfortable building this package from that branch and testing it with Android 7 to see if it fixes this issue?

mahnunchik:
Hi @MasterKale it is the strangest error I have ever seen.

It seems it happens on Android 7 devices when you **never** use FIDO auth with any value in `requireResidentKey` or `userVerification`.

If you use first time FIDO auth on **any** website with `requireResidentKey` or `userVerification` then **FIDO** is fixed for future usage.

I've tested on the following devices:

* Huawei MediaPad M3 Lite 32Gb (CPN-L09) Android 7.0
* Redmi Note 4 MIUI Global 11.0.2 Android 7.0
* Wileyfox Swift 2 Android 7.0

Preposed PR should fix this issue but I don't have more Android 7.0 devices with **not fixed** FIDO auth to test it 😢

MasterKale:
That's unfortunate that it seems to be a one-time initialization thing...who knew what Google was thinking back then 🤷‍♂️ 

Nonetheless I've merged in that update with an explicit return value for `authenticatorSelection` that will hopefully address this going forward. This is available in **@simplewebauthn/server@0.8.2**.

Thank you for reporting this :v:

PR #48, opened 2020-08-10
feature/authenticator-selection-defaults

MasterKale:
This PR adds some default values for `authenticatorSelection` to help support older platforms that may not be aware of defaults declared within the spec.

This is to address issue #47.


MasterKale:
I'm sneaking in some updated keywords for these packages' npm listings to resolve #44 with this ticket as well.

Issue #49, opened 2020-08-29
verifyAssertionResponse on TPM returns wrong tag error


yaron14:
Registration process was set with requireResidentKey: true and userVerification: "preferred".
I registered successfully using my computer PIN code but on login, after entring my PIN code I've got the following error:
error:0D0680A8:asn1 encoding routines:asn1_check_tlen:wrong tag

Am I missing something in my assertion config, or even something is missing or wrong I'm carrying from registration?

MasterKale:
Aha, I was wondering when this would come up. I've been aware of this for some time now. If I remember correctly I got as far as discovering that the public key the TPM authenticator returns during attestation is surprisingly short - maybe 16 characters? This strange "public key" causes issues with signature verification at the end of an assertion verification, which is the error you're seeing.

I've had a branch with a unit test on it that recreates this exact same error you're experiencing. I've gone ahead and drafted a WIP PR #50 to eventually contain the fix for this issue.

yaron14:
Thanks!
Waiting for your next update to install it as a package dependency.

MasterKale:
@yaron14 Can you do a Start Menu search for "TPM", click "Security Processor", and then provide a screenshot here of your "Security processor details" screen? It'll look like this:

![paste](https://user-images.githubusercontent.com/5166470/91751790-37a02100-eb7a-11ea-9f46-79acc0e031d5.png)

I'm wondering what your "Attestation" status is. If it's "not ready" like mine then it might mean there's something at the OS/BIOS level that needs to be adjusted to get it to a "Ready" state. This might be a symptom of a mis-configured system preventing valid public keys from being returned in a TPM attestation, which causes assertions to fail.

yaron14:
![image](https://user-images.githubusercontent.com/29625659/91755276-8332e900-ebd3-11ea-934f-f782241e27d8.png)

I've already tried and tested WebAuthn with some demos and libraries and my TPM authenticattor worked fine.
I believe it relates to what you've said earlier that the public key that been generated is really short and for some reason it failed the signature verification phase.

MasterKale:
I assumed SimpleWebAuthn would handle TPM fine because it passes the FIDO Conformance test which include TPM assertions, but for some reason in the real world something isn't working as expected.

Is there another library in which your same setup handles TPM assertions just fine? I'd be curious to compare notes.

Can you show me how you're calling `generateAttestationOptions()`? That'll help me recreate your conditions.

[...]

PR #50, opened 2020-08-29
feature/enhance-rsa-support

MasterKale:
This PR enhances support for attestations and assertions containing RSA authenticator public keys. The simple fact is too much code assumed it was handling EC2 public keys; going RSA public key was surprisingly underbaked.

I've renamed the PR to reflect the fact that it wasn't TPM attestation support that was fixed, but rather attestations containing RSA public keys, of which TPM devices commonly produce.

**This PR contains a breaking change!** The entire authenticator public key is now returned from `verifyAttestationResponse()` instead of a previously pared down amount of public key info. Any authenticator public keys stored before this PR are no longer valid for use in assertions.

Fortunately I was able to create a simple "upgrade" script that these "old" public keys can be run through to fix them:

https://gist.github.com/MasterKale/175cb210b097632d7cd03fd409e2dfb3


MasterKale:
Issue #49 prompted me to finally push up my initial work at tracking down this bug.

PR #51, opened 2020-09-02
feature/debugging-tools

MasterKale:
This PR includes new methods that can be used to quickly preview the contents of an attestation or assertion credential. No validation occurs, this is strictly for peering inside the contents of a response to make it easier for devs to understand what's contained within.


MasterKale:
I woke up to the realization that this will be better as a standalone browser library.

PR #60, opened 2020-10-08
fix/server-pin-dependencies

MasterKale:
I'm more aggressively pinning **server** package dependencies to exact versions because I have observed some third-party dependencies publishing minor or revision releases that actually break this package.

For example, `"@peculiar/asn1-schema": "^2.0.8"` led to `@peculiar/asn1-schema@2.0.23` being installed. Unfortunately v2.0.23 made changes to its `OctetString` class so it'd work better in ES5 code bases, which broke the `verifyAndroidKey()` and `verifyApple()` attestation methods.

I tried a plain `npm upgrade` before resorting to this but observed new versions of `jsrsasign` and/or `cbor` breaking FIDO conformance testing with ambiguous error messages that I don't have time right now to debug. The simple solution in the short term is to simply pin these packages to exact versions for more reliable installation via npm.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -32,17 +32,17 @@
     "node"
   ],
   "dependencies": {
-    "@peculiar/asn1-android": "^2.0.8",
-    "@peculiar/asn1-schema": "^2.0.8",
-    "@peculiar/asn1-x509": "^2.0.10",
+    "@peculiar/asn1-android": "2.0.8",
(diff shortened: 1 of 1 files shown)


MasterKale:
Note to self: it's probably fine to update the `@peculiar` packages to their latest release (v2.0.23), but an issue with the FIDO MDS testing server host name not resolving prevents me from ensuring FIDO conformance is preserved. For now I'll stick with these versions and upgrade later.

Issue #61, opened 2020-10-10
Error in verifyAttestationResponse when using apple format and FaceID/TouchID sometimes not showing up


P4sca1:
There is an error thrown in `verifyAssertionOptions` when using apple FaceID to solve the attestation request.

**Attestation options:**
```
{
  challenge: 'VNT_3_E2tFGe4NUoLRdJhsztnVMbI_98Vr3GgMmRVlg',
  rp: { name: 'IPS Hosting', id: 'ips-hosting.eu.ngrok.io' },
  user: { id: '21', name: 'P4sca1', displayName: 'P4sca1' },
  pubKeyCredParams: [
    { alg: -7, type: 'public-key' },
    { alg: -8, type: 'public-key' },
    { alg: -36, type: 'public-key' },
    { alg: -37, type: 'public-key' },
    { alg: -38, type: 'public-key' },
    { alg: -39, type: 'public-key' },
    { alg: -257, type: 'public-key' },
    { alg: -258, type: 'public-key' },
    { alg: -259, type: 'public-key' }
  ],
  timeout: 60000,
  attestation: 'direct',
  excludeCredentials: [],
  authenticatorSelection: { userVerification: 'discouraged' },
  extensions: undefined
}
```

[...]

PR #62, opened 2020-10-12
feature/upgrade-asn1-libs

MasterKale:
This PR updates the `@peculiar/asn1-*` libraries to their latest version v2.0.23. This should help fix part of #61 by making **server** compatible with these latest releases that are getting pulled in anyway as sub-dependencies.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.10.1",
+  "version": "0.10.2-alpha.0",
   "npmClient": "npm",
   "command": {
     "publish": {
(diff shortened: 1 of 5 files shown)


MasterKale:
@P4sca1 tested this successfully, so I'm going to go ahead and merge this.

MasterKale:
Also I acknowledge that this just kicks the can down the road a bit. It doesn't permanently solve for what to do when third-party packages release breaking changes as revision releases (v2.0.8 -> v2.0.23) when it should have been at least a minor release (v2.0.8 -> v2.1.0) ☹️

MasterKale:
For reference later PR #68 successfully restored caret use in package versions so the latest `asn1-` packages should get picked up now.

PR #67, opened 2020-11-04
chore: remove dom lib reference from server package

jstewmon:
The dom lib is incompatible with node, which can result in inaccurate
type checking. Packages which reference types provided by a lib require
package consumers to either include the lib or use skipLibCheck which
disables all type declaration checks.

So instead of referencing the dom lib, define the types needed by the
server package locally and import them where needed.


tsconfig.json
```diff
@@ -1,7 +1,10 @@
 {
   "compilerOptions": {
+    "skipLibCheck": true,
```
jstewmon, line 3: This is why I marked the PR as a draft - it won't be needed when `@peculiar/asn1-schema` is referencing a version of `@types/asn1js` that avoids the dom lib reference. That work is in progress - I hit a snag with the version number of `@types/asn1js` not being set, which precluded the updated package being picked up by the new version of `@peculiar/asn1-schema` (caret specifier with leading zeros 🤦 )

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49349
https://github.com/PeculiarVentures/asn1-schema/pull/38

MasterKale:
> The dom lib is incompatible with node...

Can you clarify this please? The DOM lib is provided by TypeScript, and types are stripped out before code ever touches Node. How can they be incompatible?

jstewmon:
The `lib` setting provides default type definitions. The setting has to be chosen based on the environment in which the code is expected to run in order for the correct set of types to be used by the type checker.

For example, the type definitions for timers e.g. `setTimeout` returns a `number` in the dom lib, but it returns `NodeJS.Timeout` in `@types/node`. There are a variety of such discrepancies that lead to programming errors or useless type narrowing checks if the `dom` lib is used in code that is intended to run outside of the browser.

see: https://www.typescriptlang.org/tsconfig#lib

jstewmon:
@MasterKale I think this PR is in good shape now. I duplicated the typedoc version bump to get the Netlify checks passing, and I used caret specifiers for the updated `@peculiar/asn1-*` deps. If there's a particular strategy you'd like to use to reconcile those changes with #68, let me know, and I'll do the needful.

MasterKale:
@jstewmon if you pull in the latest from master does that take care of the conflicts?

And I want to check my understanding of what's going on in this PR: TypeScript's `"DOM"` lib is inappropriate to include in the tsconfig.json of a NodeJS-targeted library like **server** for..._reasons_ (still not sure why, something about inaccurate type-checking)...

The fix for this scenario is to copy-paste the required DOM lib type definitions into the codebase, as you've done here in typescript-types' new **dom.ts**. The various internals of the **server** package that reference DOM lib types then need to be updated to refer to these types duplicated types.

Is this accurate?

jstewmon:
> @jstewmon if you pull in the latest from master does that take care of the conflicts?

👍 I've rebased and resolved the conflicts.

> And I want to check my understanding of what's going on in this PR: TypeScript's `"DOM"` lib is inappropriate to include in the tsconfig.json of a NodeJS-targeted library like **server** for..._reasons_ (still not sure why, something about inaccurate type-checking)...
> 
> The fix for this scenario is to copy-paste the required DOM lib type definitions into the codebase, as you've done here in **typescript-types** package's new **dom.ts**. The various internals of the **server** package that reference DOM lib types then need to be updated to refer to these duplicated types.
> 
> Is this accurate?

Yeah, you got it. The _reasons_ are about _globals_. Node and Browsers have different global APIs - both the set of available globals and the interfaces of globals whose names are available on both.

When use use a global like `setTimeout`, typescript searches for a type definition for the function. If you have the `dom` lib enabled it find [this definition](https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L19869 ). If you have `@types/node` installed, it finds [this definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/globals.d.ts#L50). If you have both, it will find both and the initial type of `setTimeout` will be overloaded with both type definitions. This can lead to the need to explicitly narrow the type with either a cast or a type assertion.

Types that are in the dom globals but not node globals can lead to programming errors like trying to instantiate a type which does not exist, as happened in https://github.com/prisma-labs/graphql-request/issues/206. The `Headers` class needed to be imported from a library, but because the DOM lib was referenced, the typescript compiler found a global type definition and the error went undetected.

There's no way to import types from a lib, so if you want to use type definitions that are already defined by the dom lib in a package that targets node, then the best solution is to copy the type definitions from the dom lib into your project.

packages/typescript-types/src/dom.ts
```diff
@@ -0,0 +1,126 @@
+export interface txAuthGenericArg {
```
MasterKale, line 1: Can you please include a block comment here noting which version of the TypeScript DOM lib these were copied from? If these are going to be copy-pasted over directly it'll be nice to version them for when the time comes that they need to be synced with a future version of TypeScript.

MasterKale (changes requested):
@jstewmon I'm satisfied with this PR save for the one comment I left in **dom.ts**. Once that's addressed this'll be ready for merge.

packages/typescript-types/src/dom.ts
MasterKale: @jstewmon Can you document here which version of the TypeScript DOM lib these were copied from? I can always add this comment myself in after this PR gets merged.

jstewmon: Sorry for the slow response... I took a few days off.

I contemplated your suggestion of documenting from which version these were lifted, and it left me feeling unsatisfied with respect to both the provenance and maintenance of these exfiltrated type definitions.

So, I came up with what I think is a nice solution, which could potentially become a model for how other projects handle this same challenge...

I've written a script using [ts-morph], which uses uses the names of the types we reference to create a dom module that exports all of the requisite interface and type aliases. It also includes a leading comment, documenting the typescript package version from which the module was generated.

I think this is a good solution because it is repeatable and precludes the tedious task of gathering all the transient type definitions by hand.

For now, I've just done enough to make the script useable, but in the future, it would be nice to have a lint step which ensures that the `dom.ts` file is unchanged by rerunning the script.

Please check it out on the new commit and LMK what you think!

[ts-morph]: https://ts-morph.com/

packages/typescript-types/src/index.ts
```diff
@@ -2,6 +2,22 @@
...
+  AuthenticatorTransport,
+  COSEAlgorithmIdentifier,
+  PublicKeyCredential,
+  PublicKeyCredentialCreationOptions,
+  PublicKeyCredentialDescriptor,
+  PublicKeyCredentialRequestOptions,
+  PublicKeyCredentialUserEntity,
+} from './dom';
+
+export * from './dom';
```
MasterKale, line 19: Is the `import type` above necessary for `export *` to work? From a maintenance perspective I can see this list of types falling out of sync with the types specified as `types` in **extract-dom-types.ts**. I think it'd be fine to over-export DOM types here in the name of reducing the number of places these types need to be specified.

packages/typescript-types/src/dom.ts
MasterKale: This is an intriguing solution. Is there any reason _not_ to add `npm run extract-dom-types` to the `build` script in **typescript-types**, just before `tsc` is called? That seems like the best place to hook in this new DOM lib extraction script so it gets reliably re-run when the package is built.

packages/typescript-types/src/index.ts
jstewmon: > Is the import type above necessary for export * to work?

No, the import necessary because those types are referenced in this file.

> From a maintenance perspective I can see this list of types falling out of sync with the types specified as types in extract-dom-types.ts.

The list of types in `extract-dom-types.ts` includes all of the types that are directly referenced in either the `typescript-types` package or the `server` package. Types that are directly referenced are listed to ensure they are included in the extraction.

If the source of confusion was just the import block in this file, does the code comment in `extract-dom-types.ts` sufficiently explain how to maintain the list?

> I think it'd be fine to over-export DOM types here in the name of reducing the number of places these types need to be specified.

`export * from ./dom` does this - it means that all of the exports in `dom.ts` are also exports of `index.ts`.

packages/typescript-types/src/dom.ts
jstewmon: I think that's a good idea. I added a new commit that adds `dom.ts` to `.gitignore` and generates the  file during the build as suggested.

MasterKale (approved):
Alright, this looks ready to go. I'll merge it in and cut a release sometime in the next couple days :shipit:

PR #71, opened 2020-11-17
feat/port-example-to-ts

MasterKale:
This PR ports the example project to TypeScript for better typing, and to prepare it for eventual deployment _somewhere_ as a demo server for people to interact with.

A new environment variable `ENABLE_CONFORMANCE` can be set to the string `"true"` to enable FIDO Conformance testing (instead of needing to comment/uncomment those couple of lines of code):

```
ENABLE_CONFORMANCE=true
```

And finally the new `npm run build` command can be used to build the example server into a typical Node server.

### DIFF
--- a/example/example-server.d.ts
+++ b/example/example-server.d.ts
@@ -1,0 +1,38 @@
+import type { AuthenticatorDevice } from '@simplewebauthn/typescript-types';
+
+/**
+ * You'll need a database to store a few things:
+ *
+ * 1. Users
+ *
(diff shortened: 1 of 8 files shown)


MasterKale:
I diagnosed the issue causing Netlify deployments to fail - they were building using Node 10 when I use at least 12. After tweaking the Netlify setup and confirming that builds are working again, I'm stuck on how to actually re-run it in this PR, so I'm going to go ahead and merge it.

PR #68, opened 2020-11-04
chore: use caret specifier for dependencies

jstewmon:
Using the caret specifier allows package consumers to adopt newer,
compatible versions of transient dependencies, which has the following
benefits:
- de-duplication of compatible package versions
- bug and security fixes can be adopted without waiting on a new version
  of the referencing package

--------

👋  Hi, I found this to be an issue when reinstalling `@simplewebauthn/server` in my project did not pick up the latest version of `@peculiar/asn1-schema` as expected.

PR #72, opened 2020-11-17
fix: add missing type defs to example package.json

jstewmon:


MasterKale:
@jstewmon Can I trouble you to also include `@types/node` in this PR as well? I think types in the monorepo's package.json are being picked up when I run any `npm` commands within **example/**, and when I isolate this project into an external folder and try to build it TypeScript complains about the missing Node types as well as the other two you're adding here.

jstewmon:
Sure, no problem. Which version of node do you want to target? The major version of `@types/node` should correspond to the node major version.

MasterKale:
CI tests 12 and 14, so let's go with 14 for now since this is an example server and it can be a little more liberal with its support for the latest and greatest.

MasterKale:
@jstewmon I fixed the Netlify deployment issue with PR #74. If you pull in `master` into your branch the docs build step should start passing again. #67 and #68 will need that merge too.

jstewmon:
@MasterKale I've rebased all three PRs 👍

MasterKale (approved):

PR #79, opened 2020-12-04
Optional Allow Credential

Moumouls:
`allowCredentials` can be optional: https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialRequestOptions/allowCredentials

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,3 +49,9 @@ Running Jest in watch mode for a specific project requires the use of `lerna exe
 ```sh
 $> npx lerna exec npm run test:watch --scope=@simplewebauthn/server
 ```
+
+Dev on the server package with jest watch target
+
+```sh
+$> npm run dev:server aFileNameHere.js
+```
--- a/README.md
+++ b/README.md
@@ -50,8 +50,3 @@ Running Jest in watch mode for a specific project requires the use of `lerna exe
 $> npx lerna exec npm run test:watch --scope=@simplewebauthn/server
 ```
 
-Dev on the server package with jest watch target
-
-```sh
-$> npm run dev:server aFileNameHere.js
-```
--- a/packages/browser/src/methods/startAssertion.test.ts
+++ b/packages/browser/src/methods/startAssertion.test.ts
@@ -39,6 +39,12 @@ const goodOpts2UTF8: PublicKeyCredentialRequestOptionsJSON = {
   timeout: 1,
 };
 
+// Without allow credentials
+const goodOpts3: PublicKeyCredentialRequestOptionsJSON = {
+  challenge: bufferToBase64URLString(toUint8Array('fizz')),
+  timeout: 1,
+};
+
 beforeEach(() => {
   mockNavigatorGet.mockReset();
   mockSupportsWebauthn.mockReset();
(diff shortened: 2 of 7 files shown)


Moumouls:
@MasterKale not sure about what happening into the CI, `npm test` is okay on my side, TS is also ok on the file. Do you have an idea ?

MasterKale:
@Moumouls I looked at Actions output and TypeScript is complaining that the `PublicKeyCredentialRequestOptionsJSON` type in the **typescript-types** package wasn't updated to allow `allowCredentials` to be optional. I don't know why it's compiling on your end but it's a legitimate issue with this PR.

I'll review the rest of the PR later today.

Moumouls:
@MasterKale okay i understand the issue, since i'm not sure to understand the typescript types package, is there any way to add the typecheck of this package via `npm test`. Currently error is only raised after a full `bootstrap` command.

packages/browser/src/methods/startAssertion.test.ts
```diff
@@ -59,15 +65,20 @@ test('should convert options before passing to navigator.credentials.get(...)',
     },
   );
 
-  await startAssertion(goodOpts1);
+  const checkWithOpts = async (opts: PublicKeyCredentialRequestOptionsJSON) => {
```
MasterKale, line 68: Can you please create a new unit test for "should support optional allowCredentials" instead of expanding this existing one?

packages/browser/src/methods/startAssertion.ts
```diff
@@ -25,7 +25,12 @@ export default async function startAssertion(
   const publicKey: PublicKeyCredentialRequestOptions = {
     ...requestOptionsJSON,
     challenge: base64URLStringToBuffer(requestOptionsJSON.challenge),
-    allowCredentials: requestOptionsJSON.allowCredentials.map(toPublicKeyCredentialDescriptor),
+    // We need to avoid passing empty array to avoid blocking retrieval
+    // of public key
+    allowCredentials:
+      requestOptionsJSON.allowCredentials?.length === 0
```
MasterKale, line 31: I'm not a huge fan of ternaries - can you please move this logic to a separate `if ()` statement below this definition of `publicKey`?

MasterKale (changes requested):

README.md
```diff
@@ -49,3 +49,9 @@ Running Jest in watch mode for a specific project requires the use of `lerna exe
 ```sh
 $> npx lerna exec npm run test:watch --scope=@simplewebauthn/server
 ```
+
+Dev on the server package with jest watch target
```
MasterKale, line 53: This addition to the README may not be needed since the Jest watch mode offers much more robust ways of narrowing tests by test name, filename, etc... Can we leave this out for that functionality?

Moumouls: it was just for DX, the suggested `npx lerna exec npm run test:watch --scope=@simplewebauthn/server` i think is a little bit too long. You just want to remove it from the readme ?

Moumouls: it was just for DX, the suggested npx lerna exec npm run test:watch --scope=@simplewebauthn/server i think is a little bit too long. You just want to remove it from the readme ?

packages/browser/src/methods/startAssertion.ts
Moumouls: May be te ternary is inverted and should finish by `undefined`,but i think it's here the good way to achieve the result.
What bothers you with ternary ? :)

README.md
MasterKale: I think the new `dev:server` and `dev:browser` scripts are great, this comment is specifically about removing this addition to the README.

packages/browser/src/methods/startAssertion.test.ts
```diff
@@ -72,6 +84,31 @@ test('should convert options before passing to navigator.credentials.get(...)',
...
+        });
+      });
+    },
+  );
+
+  await startAssertion(goodOpts3);
+  let allowCredentials = mockNavigatorGet.mock.calls[0][0].allowCredentials;
+  expect(allowCredentials).toEqual(undefined);
+
+  // Should convert empty array to undefined
```
MasterKale, line 106: This should become a separate unit test too.

packages/browser/src/methods/startAssertion.test.ts
```diff
@@ -72,6 +84,31 @@ test('should convert options before passing to navigator.credentials.get(...)',
...
+      return new Promise(resolve => {
+        resolve({
+          response: {},
+          getClientExtensionResults: () => ({}),
+        });
+      });
+    },
+  );
+
+  await startAssertion(goodOpts3);
```
MasterKale, line 102: `goodOpts3` isn't used anywhere but this test, can you please move its definition into here instead? Same thing with `goodOpts4` when you split that into a new unit test, move its declaration into the unit test so we have fewer "goodOpts" variables (it's a bad pattern I established early on, I'm trying to prevent it from becoming the norm 😅)

MasterKale (changes requested):

packages/server/src/assertion/generateAssertionOptions.ts
```diff
@@ -20,7 +20,8 @@ type Options = {
 /**
  * Prepare a value to pass into navigator.credentials.get(...) for authenticator "login"
  *
- * @param allowCredentials Authenticators previously registered by the user
+ * @param allowCredentials Authenticators previously registered by the user, if not provided
+ * device can ask user which credential he wants to use
```
MasterKale, line 24: This is a slight docs tweak but can you make it say this instead?

> Authenticators previously registered by the user, if any. If undefined the client will ask the user which credential they want to use

MasterKale (approved):
This looks great, thanks for including unit tests :v:

PR #80, opened 2020-12-04
Adapter sytem + JWT Adapter

Moumouls:


akanass (commented):
The principle of the challenge in security is to not be transferred in the network to be a real proof of trust and defend against session replay attacks like explain in this [article](https://searchsecurity.techtarget.com/definition/challenge-response-system) 

So for me I disagree this PR

Moumouls:
@MasterKale  since it's an optional feature, only here for better DX if developers don't want to store challenge into the DB let me know if it stay interesting to keep the option.

Moumouls:
@akanass may be there is also something that i miss here, may be you can help me to understand details :)  
`the principle of the challenge in security is to not be transferred in the network`

But currently Webauthn spec and the current simple web authn transfer the challenge through the network.
If i'm not wrong, challenge in WebAuthN spec is only used to avoid replay attacks and ensure that user currently have the private key.

If i understand the risk could be that an attacker can replay attack X times within the JWT time validity ?

akanass:
@Moumouls is like for `OAuth2` or `OpenID Connect` protocol.

Moumouls:
https://stackoverflow.com/a/2770135 is pretty interesting and if i understand it allow replaying an opaque action(ex: send again 100$) without carrying about the returned value since the returned value is opaque due to SSL.

In auth user story (like webauthn), the main target is to obtain a session token return after server assertion check.

With a JWT expiration set same as the webauthn timeout option (default 60sec).

As an attacker: 
- i can sniff the network
- i can find and intercept opaque https assertion request
- i can replay the assertion request during the JWT validity time (ex: 60sec), if assertion is ok, server will return new session token at each replay (during remaining JWT time), but since response is through SSL, attacker will not get access to the session token and can't log in

Here i can see drawback similar to DB implementation:
- On DB implementation: an attacker can replay as he wants a challenge request (and the server will store many and many challenge into DB, if any system start to mitigate challenge generation)
- On JWT: an attacker can replay as he wants a challenge request but no challenge is stored, (just signed). But during the JWT validity the attacker can generate X tokens 

Sorry @akanass all of this comments :), i'm trying to dive deeply into the concept, to identify plausible usable breach and real risks of this implementation.

akanass:
@Moumouls your link is the same as mine in my answer :)

In DB/cache implementation, you have to check, before creating a challenge request, if one is already existing to avoid the possibility to ask many one for the same user.

In Yubico server implementation they are doing that.

Like this you prevent all bad intention or bad implementation in your front-end application and a user clicking and clicking on the registration button.

Like said in my previous answer, `OAuth` and `OpenID Connect` protocols are using that and they are the most famous authentication process.

You can see a simple implementation from `Fido` [here](https://github.com/fido-alliance/webauthn-demo/tree/completed-demo) and you will see they are doing like this and they are the bosses.

Moumouls:
I understand @akanass you opinion, but based on the risk evaluation done here, I think we fall into the famous debate similar to : JWT or DB Session token?

i think there are not correct answer since these 2 systems have pro/cons. 

My opinion here is to let the developers choose a `Challenge Storage`, "DB one" or "JWT one" depending of it needs.

i can introduce a feature like this with default transport to "DB".

akanass:
@Moumouls the developer should choose what the standard are doing and specially what official and creator of this standard are doing, it's my point of view

Moumouls:
So based on your last comment @akanass and also my experience from https://github.com/parse-community/parse-server. I introduced an Adapter system to allow developers to customize main functions output/input.

I'have implemented the JWTChallengeAdapter as an example. Then developers will a have some flexibility to validate/modify input/output, and create their own adapters.

This way the main interface will stay standard, since all custom logic wrapped into an optional `adapters` key

If the Adapter addition looks good to you @MasterKale / @akanass but you don't want the JWT Challenge Adapter on this project. I can remove it.... and i will copy the code to parse-server core code, since JWT Challenge Adapter is the perfect solution for Parse Platform community (https://parseplatform.org/); the implementation will be fast, and do not pollute DB.

packages/server/src/adapters/JWTChallengeAdapter.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export interface SignChallengePayload {
+  challenge: string;
+  rpID: string;
+  origin: string;
+}
+
+export default class JWTChallengeAdapter extends BaseAdapter {
+  secret: string;
```
akanass, line 18: If you want to implement an adapter, you should allow all options for `JWT` creation and using a simple secret it's not the most secure way to do it.

I'm always using `RSA` key/pair when implementing a `JWT` solution to avoir someone to find your secret.

akanass (changes requested):

packages/server/src/adapters/JWTChallengeAdapter.ts
akanass: And like @MasterKale said before, I think this solution is most in a server logic than inside this library

akanass:
And like @MasterKale said before, I think this solution is most in a server logic than inside this library

packages/server/src/adapters/JWTChallengeAdapter.ts
Moumouls: You use something like that ?

```
var token = jwt.sign({ foo: 'bar' }, privateKey, { algorithm: 'RS256'})
```

packages/server/src/adapters/JWTChallengeAdapter.ts
```diff
@@ -49,15 +53,17 @@ export default class JWTChallengeAdapter extends BaseAdapter {
...
   }
 
   verifyChallenge(opts: verifyAssertIO | verifyAttestIO): void {
     const response = opts.credential.adapters?.[this.key];
     if (!response) super.throwMissingKey();
 
-    const signedChallengePayload = verify(response, this.secret) as SignChallengePayload;
+    const signedChallengePayload = verify(response, this.secret, {
+      algorithms: [this.jwtAlgorithm],
+    }) as SignChallengePayload;
```
akanass, line 66: ALL [options](https://github.com/auth0/node-jsonwebtoken#jwtverifytoken-secretorpublickey-options-callback) for a `JWT` verification should be allowed

packages/server/src/adapters/JWTChallengeAdapter.ts
```diff
@@ -49,15 +53,17 @@ export default class JWTChallengeAdapter extends BaseAdapter {
         origin: this.options.origin,
```
akanass, line 53: ALL [options](https://github.com/auth0/node-jsonwebtoken#jwtsignpayload-secretorprivatekey-options-callback) for a `JWT` signature should be allowed

akanass (changes requested):

packages/server/src/adapters/JWTChallengeAdapter.ts
Moumouls: yeah a type extract from jsonwebtoken @types package should be the right way

Moumouls:
@akanass  also with the adapter system we could imagine a RedisAdapter for DB challenge storage 😉  for an easier usage for users that already have a Redis deployed, then this package could be even more plug and play :)

akanass:
@Moumouls adapter system is the best way to do it but we have to provide a generic interface to be implemented and used inside this library.

If no adapter is provided/implemented, the basic one is used.

This solution will be the best one.

Moumouls:
the generic interface is currently `BaseAdapter`. And could be easly improved over time :)

akanass:
@Moumouls I agree but we have to give an `Abstract Class` with all methods to implement and which are used in the library to override the basic one.

Like this user can manage it's own connexion to a dedicated DB or manage the challenge in a JWT etc without breaking the flow

akanass:
Like explain in the implementation of [Yubico server](https://github.com/Yubico/java-webauthn-server/tree/master/webauthn-server-demo),
```
This layer manages the general architecture of the system, and is where most business logic and integration code would go. The demo server implements the "persistent" storage of users and credential registrations - the CredentialRepository integration point - as the InMemoryRegistrationStorage class, which simply keeps them stored in memory for a limited time. The transient storage of pending challenges is also kept in memory, but for a shorter duration.
```

Moumouls:
@akanass tell me if you see something that we can improve.

packages/server/src/adapters/JWTChallengeAdapter.ts
```diff
@@ -0,0 +1,91 @@
...
+    super();
+    const { secretOrPrivateKey, jwtOptions } = options;
+    this.options = options;
+    if (secretOrPrivateKey && secretOrPrivateKey.length < this.recommendedSecretLength) {
+      throw new Error(
+        `jwt secret seems too weak please use a secret with more than ${this.recommendedSecretLength} chars`,
+      );
+    }
+    this.secret = secretOrPrivateKey;
+    if (jwtOptions) this.jwtOptions = jwtOptions;
```
akanass, line 41: it should be better to do `Object.assign({}, this.jwtOptions, jwtOptions)` to be sure default options are still present if not provided.

I think it's better to separate `sign` and `verify` options to be sure to override properly and having default one as well if not provided

packages/server/src/assertion/testHelper.ts
```diff
@@ -0,0 +1,49 @@
+/**
```
akanass, line 1: should not create a file like this because it'll be in final package - missing `.test.ts` extension

packages/server/src/attestation/testHelper.ts
```diff
@@ -0,0 +1,51 @@
+import { PublicKeyCredentialCreationOptionsJSON } from '@simplewebauthn/typescript-types';
```
akanass, line 1: same remark here like for `assertion` one, this file should not exist

packages/server/src/helpers/reducePromise.ts
```diff
@@ -0,0 +1,12 @@
+export default async function reducePromise<K, T>(
```
akanass, line 1: not a big fan to put `Promise` everywhere if not needed and this custom function change to much the way to process in the whole library

packages/server/src/adapters/JWTChallengeAdapter.ts
```diff
@@ -0,0 +1,91 @@
...
+      this.secret,
+      { expiresIn: '2m', ...this.jwtOptions.sign },
+    );
+  }
+
+  verifyChallenge(opts: verifyAssertIO | verifyAttestIO): void {
+    const response = opts.credential.adapters?.[this.key];
+    if (!response) super.throwMissingKey();
+
+    const signedChallengePayload = verify(
```
akanass, line 67: `JWT.verify` should validate more fields to be secure like explain [here](https://curity.io/resources/architect/api-security/jwt-best-practices/)

akanass (changes requested):
I put some comments and I don't feel totally confortable with all this implementation, it seems to much personal for me, we can feel your own implementation style and for a library it’s to risky in my point of view.

MasterKale:
@Moumouls I appreciate the effort you've put into this. I think it's an interesting solution to a problem I hadn't considered, and at a glance your solution seems to be a good one in tackling your particular use case.

That said, functionality in the **server** and **browser** packages need to remain scoped to the WebAuthn spec to make it easier to evolve this project alongside the spec. I don't think the changes in this PR are appropriate for this project as its currently organized. _Maybe_ this becomes the start of a new "@simplewebauthn/express" package to contain opinionated, implementation-specific code like this. Until the spec dictates methods of transferring the challenge to a client, though, I think PR's like this fall outside of the scope of this project.

Moumouls:
Okay, thanks for your feedbacks @akanass @MasterKale , i understand out of scope issue. Sadly I'm not a huge fan of my last option but i think i'm gonna to start a fork under Parse Organization and then re publish the package for Parse community developers. 
I will check my options :) 

Let's close this one !

akanass:
I think we can all work together to provide full complete server implementation like `Yubico` is doing.

I'm currently working on a server in `NestJS` an using this library so after that I can extract some code to have an easy way to implement a server and we can work hand to hand on it and publish it under this repository :)

MasterKale:
> We can work hand to hand on it and publish it under this repository :)
I want to reiterate that I'm not at all against the creation of a new "@simplewebauthn/express" package within this project (I'm using Express for sake of discussion, it could be something general purpose or maybe this project does offer multiple integration support packages for Express/Parse/NestJS/etc...).

@Moumouls If this idea appeals to you, perhaps this PR can be re-architected into a new "@simplewebauthn/parse" package? I see the value in your code, I just think the **server** and **browser** packages need to remain simple so they can adhere to the spec but also work as foundational packages for Node server-specific frameworks.

Moumouls:
Following the YAGNI methodology (that I particularly like), parse server/JS SDK implementation only need at this time the JWT system, i will go for a customized implementation in [parse-server](https://github.com/parse-community/parse-server) code using the basic `@simplewebauthn/server`. I really want to avoid the fork.

`@simplewebauthn/parse` is a good idea but a `webauthn-adapter` package should be better under under parse organization. Currently Auth adapters are super light weight a package will be overkill 😉 I'm gonna to onboard directly the WebAuth business logic (with your browser/server package) into `parse-server` and  `parse` packages to give a plug and play experience to our [parse](https://parseplatform.org/) community 👍

akanass:
@Moumouls I found something interesting in the official specification saying the challenge should be store in the RP server and not coming to the client side https://www.w3.org/TR/webauthn/#cryptographic-challenges

Moumouls:
@akanass the challenge comes every time to client, I think they talk about the fact that the client cannot generate the challenge and the server should be in capacity to check that the singed challenge match a challenge iniated by the server whatever the persistence method (db, jwt, etc...)

Issue #82, opened 2020-12-08
Export some test utils


Moumouls:
For an easier server integration testing using this lib i think it could be helpful to export some assertion/attestation examples (output of startAssertion/startAttestation).
What do you think about that ?

MasterKale:
That's an interesting proposal. This could involve moving the various sample attestations and assertions used in the various .test.ts files into a separate testing-oriented package 🤔

Moumouls:
Yes testing package will be the right solution !
Here i only first think about the `server` integration testing part

But we can imagine later a plug and play testing util function for E2E front testing (ex: for cypress)

PR #84, opened 2020-12-10
feat/testing-package

MasterKale:
This PR is in response to a request in #82. This adds a new "testing" package to the project.

Going forward the plan is to relocate many of the credentials that are scattered throughout the various unit tests in the **server** and **browser** packages into **testing**. They'll eventually get combined with various configurations of attestation and assertion options so that projects implementing SimpleWebAuthn can import these values from **testing** to use in their own unit and integration tests.

Migrating values into **testing** won't happen overnight. This will be a gradual migration as future work is completed on this project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,11 +2,12 @@
   "name": "simplewebauthn-monorepo",
   "private": true,
   "scripts": {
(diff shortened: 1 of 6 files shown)


Moumouls:
Perfect, I think I could push some PR next week :)

PR #85, opened 2020-12-15
Testing package utils

Moumouls:
- [x] Architecture
- [x] Assertion Utils
- [x] Attestation Utils
- [x] Tests
- [x] Data consistency

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "simplewebauthn-monorepo",
   "private": true,
   "scripts": {
-    "bootstrap": "npm run build:types && npm run build:testing && npm run build:browser && npm run build:server",
+    "bootstrap": "npm run build:types && npm run build:browser && npm run build:server && npm run build:testing",
     "lint": "prettier --write packages/ example/ && eslint --fix packages/ example/",
     "docs": "npm run bootstrap && typedoc --tsconfig tsconfigdoc.json",
     "test": "lerna run test",
--- a/packages/server/src/assertion/generateAssertionOptions.ts
+++ b/packages/server/src/assertion/generateAssertionOptions.ts
@@ -8,7 +8,7 @@ import base64url from 'base64url';
 
 import generateChallenge from '../helpers/generateChallenge';
 
-type Options = {
+export type GenerateAssertionOptions = {
   allowCredentials?: PublicKeyCredentialDescriptorJSON[];
   challenge?: string | Buffer;
   timeout?: number;
@@ -31,7 +31,7 @@ type Options = {
  * @param rpID Valid domain name (after `https://`)
  */
 export default function generateAssertionOptions(
-  options: Options = {},
+  options: GenerateAssertionOptions = {},
 ): PublicKeyCredentialRequestOptionsJSON {
   const {
     allowCredentials,
--- a/packages/server/src/assertion/verifyAssertionResponse.ts
+++ b/packages/server/src/assertion/verifyAssertionResponse.ts
@@ -12,7 +12,7 @@ import verifySignature from '../helpers/verifySignature';
 import parseAuthenticatorData from '../helpers/parseAuthenticatorData';
 import isBase64URLString from '../helpers/isBase64URLString';
 
-type Options = {
+export type VerifyAssertionResponseOptions = {
   credential: AssertionCredentialJSON;
   expectedChallenge: string;
   expectedOrigin: string;
@@ -36,7 +36,9 @@ type Options = {
  * `generateAssertionOptions()`. Activates FIDO-specific user presence and verification checks.
  * Omitting this value defaults verification to a WebAuthn-specific user presence requirement.
  */
-export default function verifyAssertionResponse(options: Options): VerifiedAssertion {
+export default function verifyAssertionResponse(
+  options: VerifyAssertionResponseOptions,
+): VerifiedAssertion {
   const {
     credential,
     expectedChallenge,
--- a/packages/server/src/attestation/generateAttestationOptions.ts
+++ b/packages/server/src/attestation/generateAttestationOptions.ts
@@ -11,7 +11,7 @@ import base64url from 'base64url';
 
 import generateChallenge from '../helpers/generateChallenge';
 
-type Options = {
+export type GenerateAttestationOptions = {
   rpName: string;
   rpID: string;
   userID: string;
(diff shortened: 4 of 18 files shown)


Moumouls:
Ready for review @MasterKale  👍

MasterKale:
Please include a write-up in the PR description of the architecture of your changes so that I can review them with your intentions in mind.

Right now I'm fine with some of the changes, like renaming and exporting the `Options` types for the various exports in **server**, but I disagree with making **server** a dependency of **testing** and testing the former within the latter (unit tests for server should reside within server so you're not looking into two places for coverage). I'm also not sure why anything is a Class with static methods then they could just as easily be simple functions that return an object, or even just the exported object.

Moumouls:
PR Goal: export some valid combination of input/output of the server packages to help developers to add some unit/integration tests easily on their projects depending of their needs.

Hi @MasterKale thanks for the review.

You just confirm my doubt about using class, i think here the solution is export function and also a default since exporting juste object could lead to unintended object mutation.

About the server `dependency` since it's just used for exported types, what do you think about moving Options type to the typescript-types package ?

If i understand you are not in favor of using `server` functions into `testing` unit tests ? I used this strategy to ensure that testing package input/outputs are functionally consistent across packages since TS can't ensure this. Do you have a suggestion or an idea here about the ideal testing method ? 

Summary
- [ ] Change class to function, with `export` and a `export default`

MasterKale:
Apologies for the delay, real life got hectic. I finally took another look at your code and now understand what you're doing in files like **packages/testing/src/attestation/verifyAttestationResponse.test.ts**: you're testing that the expected `result`s match up with what the methods in the **server** package return for the corresponding credential.

First off, I like how you pair `options` with the expected `result`, to communicate what a given response should get back from verification. The more I think about it, though, the more I think that instances of "test data" within the **testing** package should be simple objects structured like this (an example I cobbled together from a test in **verifyAssertionResponse.test.ts** combined with your renaming of the various `Options`):

```ts
import type { VerifyAssertionOptions, VerifiedAssertion } from '@simplewebauth/server';

type AssertionVerificationTestData = {
  options: VerifyAssertionOptions;
  result: VerifiedAssertion;
}

export const assertionFirstTimeUsed: AssertionVerificationTestData = {
  options: {
    credential: {
      id: 'wSisR0_4hlzw3Y1tj4uNwwifIhRa-ZxWJwWbnfror0pVK9qPdBPO5pW3gasPqn6wXHb0LNhXB_IrA1nFoSQJ9A',
      rawId: 'wSisR0_4hlzw3Y1tj4uNwwifIhRa-ZxWJwWbnfror0pVK9qPdBPO5pW3gasPqn6wXHb0LNhXB_IrA1nFoSQJ9A',
      response: {
        authenticatorData: 'PdxHEOnAiLIp26idVjIguzn3Ipr_RlsKZWsa-5qK-KABAAAAAA',
        clientDataJSON:
          'eyJjaGFsbGVuZ2UiOiJkRzkwWVd4c2VWVnVhWEYxWlZaaGJIVmxSWFpsY25sQmMzTmxjblJwYjI0IiwiY2xpZW50RXh0ZW5zaW9ucyI6e30sImhhc2hBbGdvcml0aG0iOiJTSEEtMjU2Iiwib3JpZ2luIjoiaHR0cHM6Ly9kZXYuZG9udG5lZWRhLnB3IiwidHlwZSI6IndlYmF1dGhuLmdldCJ9',
        signature:
          'MEQCIBu6M-DGzu1O8iocGHEj0UaAZm0HmxTeRIE6-nS3_CPjAiBDsmIzy5sacYwwzgpXqfwRt_2vl5yiQZ_OAqWJQBGVsQ',
      },
      type: 'public-key',
    },
    expectedChallenge: base64url.encode('totallyUniqueValueEveryAssertion'),
    expectedOrigin: 'https://dev.dontneeda.pw',
    expectedRPID: 'dev.dontneeda.pw',
    authenticator: {
      publicKey:
        'pQECAyYgASFYIGmaxR4mBbukc2QhtW2ldhAAd555r-ljlGQN8MbcTnPPIlgg9CyUlE-0AB2fbzZbNgBvJuRa7r6o2jPphOmtyNPR_kY',
      credentialID:
        'wSisR0_4hlzw3Y1tj4uNwwifIhRa-ZxWJwWbnfror0pVK9qPdBPO5pW3gasPqn6wXHb0LNhXB_IrA1nFoSQJ9A',
      counter: 0,
    },
  },
  result: {
    verified: true,
    authenticatorInfo: {
      counter: 0,
      base64CredentialID:
        'wSisR0_4hlzw3Y1tj4uNwwifIhRa-ZxWJwWbnfror0pVK9qPdBPO5pW3gasPqn6wXHb0LNhXB_IrA1nFoSQJ9A',
    }
  },
};
```

Then, the corresponding test case in **server** gets refactored into this:

```ts
import { assertionFirstTimeUsed } from '@simplewebauthn/testing';

test('should not compare counters if both are 0', () => {
  const verification = verifyAssertionResponse(assertionFirstTimeUsed.options);

  expect(verification.verified).toEqual(assertionFirstTimeUsed.result.verified);
});
```

This way **testing** contains no unit tests, no classes, just a bunch of attestation and assertion inputs and outputs, at which point it's up to the test cases in **server** (or any other implementing codebase) like yours to feed credentials through your code that's wrapping the SimpleWebAuthn methods to make sure your code is functioning as expected.

I also realize that my example justifies your importing **server** into **testing** by typing `options` and `result` with `VerifyAssertionOptions` and `VerifiedAssertion` respectively, since both of these would come from **server**. It doesn't feel right to move these types into **typescript-types** so I now have one less objection to your PR; that is, I think it's fine for **testing** to import  **server**, so long as it doesn't lead to some kind of crazy infinite import loop in **server**'s unit tests 🤔

Moumouls:
No problem @MasterKale 😃 

Thanks for your feedbacks, i have completed parse server tests.(https://github.com/parse-community/parse-server/pull/7079/commits/df69863f8346f5fedb5ffb1e3bed44ac96c8fb10)

So i exactly see what we should have here for an easier testing. 

Node JS support cyclic deps but here to reduce the cyclic behavior, we can try to read types directly from the file with something like this: `import type { VerifyAssertionOptions } from '@simplewebauth/server/dist/assertion/verifyAssertion';`

Then to avoid object mutation and break some testing isolation rules, we should use functions : `assertionFirstTimeUsed()`

And i totally agree with importing testing into server to avoid dual source of truth. Moreover, it will be a good proof of the usefulness of the test package. It will also serve as a document since the developers who will check the source code of the tests, will see the best practices and the use of the test package.

Mikescops:
I really like the idea of this PR, I went through the same process of adding integration testing on my projects and that could be much cleaner with this.
Hope you'll manage to finish this PR @Moumouls 👍

Moumouls:
Hi @Mikescops , I hope I can finish this one in 2-3 weeks ☺️

MasterKale:
It's been a while with no activity so I'm going to close this for now. @Moumouls I still think is a noble endeavor worth pursuing; if you decide to take up the mantle again please feel free to reopen this PR 👑

PR #87, opened 2020-12-21
Lazy Text Encoder

Moumouls:
Found an issue during some JSDOM Jest tests.

TextEncoder is not currently supported correctly into `jsdom` env of Jest.

To avoid a systemic crash at import time it's better to init text encoder into the function.

A workaround is to use the `node` env of Jest. But some tests require jsdom and could import indirectly the package.

### DIFF
--- a/packages/browser/src/helpers/toUint8Array.ts
+++ b/packages/browser/src/helpers/toUint8Array.ts
@@ -1,9 +1,7 @@
-const utf8Encoder = new TextEncoder();
-
 /**
  * A helper method to convert an arbitrary string sent from the server to a Uint8Array the
  * authenticator will expect.
  */
 export default function toUint8Array(value: string): Uint8Array {
-  return utf8Encoder.encode(value);
+  return new TextEncoder().encode(value);
 }


MasterKale:
I originally fixed this with **browser**'s custom [jest-environment.js](https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/browser/jest-environment.js). Is this no longer working? Unit tests are still passing in this library, so is this PR to fix issues with your Jest setup?

Moumouls:
> I originally fixed this with **browser**'s custom [jest-environment.js](https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/browser/jest-environment.js). Is this no longer working? Unit tests are still passing in this library, so is this PR to fix issues with your Jest setup?

In this library it works, but not in other projects (e.g. Parse SDK JS). This error happened in Parse SDK JS unit tests. The solution with custom jest env will work, but it seems to me more appropriate also to avoid this error by just doing an import 😃 
Also i don't know when jest will add the support of TextEncoder into the JSdom.

MasterKale:
@Moumouls I don't know why I resisted this for so long - thank you for helping make this library more pleasant to integrate 😅

MasterKale:
This has been published to NPM as **browser@0.10.6**

PR #91, opened 2021-01-21
feature/multiple-origins

### LINKED ISSUE #90: Multiple expectedOrigin and expectedRPID
Hello,

Thanks for your great work.

I have a use case that seems not working with the current implementation of the lib.
When using SimpleWebAuthn on a browser extension I have different expectedOrigin and expectedRPID for Firefox / Chrome / ...
Maybe I'm wrong but the only way I see to have the lib working with this use case is to change both `expectedOrigin` and `expectedRPID` from string to tables so that I can check if the origin and rpId in the attestation/assertion are in the list.

Is my interpretation correct and is it worth to support this use case in the lib?

MasterKale:
This PR adds support for multiple origins and RP IDs to `verifyAttestationResponse()` and `verifyAssertionResponse()`. In addition to a single string value, an array of values can now be provided:

```js
const verification = await verifyAttestationResponse({
  // ...
  expectedOrigin: ['https://simplewebauthn.dev', 'https://fizz.buzz'],
  expectedRPID: ['simplewebauthn.dev', 'fizz.buzz'],
  // ...
});
```

```js
const verification = verifyAssertionResponse({
  // ...
  expectedOrigin: ['https://simplewebauthn.dev', 'https://fizz.buzz'],
  expectedRPID: ['simplewebauthn.dev', 'fizz.buzz'],
  // ...
});
```

This should resolve #90.


packages/server/src/assertion/verifyAssertionResponse.ts
```diff
@@ -87,8 +87,16 @@ export default function verifyAssertionResponse(options: Options): VerifiedAsser
   }
 
   // Check that the origin is our site
-  if (origin !== expectedOrigin) {
-    throw new Error(`Unexpected assertion origin "${origin}", expected "${expectedOrigin}"`);
+  if (Array.isArray(expectedOrigin)) {
+    if (!expectedOrigin.includes(origin)) {
```
Mikescops, line 91:
```suggestion
  if (Array.isArray(expectedOrigin) && !expectedOrigin.includes(origin)) {
```

Mikescops (approved):
Sounds good, just added a small comment.

packages/server/src/assertion/verifyAssertionResponse.ts
MasterKale: I agree that the nested IF statement looks a little clunky, but I purposefully went with it for the sake of better type guarding for TypeScript. Nested, TypeScript knows that `expectedOrigin` in the `else` branch is always going to be a string:

Screen Shot 2021-01-21 at 1 58 39 PM

If I collapse the `if ()` statement  as suggested above then TypeScript (correctly) infers that `expectedOrigin` could still be an array in the `else` branch:

Screen Shot 2021-01-21 at 1 58 57 PM

Mikescops: Ok right, true true! I didn't thought about it 😅

MasterKale:
These changes have been published as part of the latest **v1.0.0 release (🎉)**

Mikescops:
Thanks a lot @MasterKale, I'll update my dependencies tomorrow morning first hour 😄 🚀

Issue #96, opened 2021-02-01
How to use usernameless and passwordless at the same time


akanass:
As described in the [Yubico developer Guide](https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/User_Handle.html) we can use `User Handle` to allow usernameless and passwordless authentication.

How to do it with this library and initiate `attestation` and `assertion` phases without a `username` because all documentation describes an associated user passed in parameter.

Thanks for your help.

akanass:
I think I have my answer :)

If we don't give a `username` during `assertion phase` the client will ask the user which one he wants to use but to validate it correctly after, we have to store `User Handle` and use it like describe in the documentation.

This validation should be added inside `assertion verification phase` because for now only the `string` type is [checked](https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/server/src/assertion/verifyAssertionResponse.ts#L110) for this attribute.

And to store `User Handle` we have to ask the user to use the key ones before logout like a validation process because `User Handle` is not returned during `attestation verification phase`

akanass:
https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/Resident_Keys.html

akanass:
If we want to authenticate a user with his `username` we have to set [residentKey](https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/server/src/attestation/generateAttestationOptions.ts#L128) to `required` as describe in the [specification](https://www.w3.org/TR/webauthn/#client-side-discoverable-credential)

Else, we have to ask him his `username`, retrieve authenticators associated to it, asking for user verification and then validating `User Handle` to be sure all is good

MasterKale:
passwordless.dev is @aseigler's project and my goto for playing around with Passwordless and Usernameless. I took a look at attestation options and distilled the differences down to this:

## Passwordless
```json
{
  "user": {
    "name": "namehere",
    "id": "bmFtZWhlcmU",
    "displayName": null
  },
  "authenticatorSelection": {
    "requireResidentKey": false,
    "userVerification": "preferred"
  }
}
```

## Usernameless
```json
{
  "user": {
    "name": " (Usernameless user created at 2/1/2021 5:53:46 PM)",
    "id": "IChVc2VybmFtZWxlc3MgdXNlciBjcmVhdGVkIGF0IDIvMS8yMDIxIDU6NTM6NDYgUE0p",
    "displayName": null
  },
[...]

Issue #95, opened 2021-02-01
Missing stored data in server


akanass:
As described in the [Yubico developer Guide](https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/Registering_Multiple_Devices.html) to backing up authenticator data, `userHandle`, `metadata`, and `attestation` data should be stored in the database.

For the `attestation`, see my [comment](https://github.com/MasterKale/SimpleWebAuthn/issues/95#issuecomment-771060044)

For the `userHandle`, the value exists and it's the `user.id` field during `attestation generation phase` - we have to store it during `attestation verification phase` inside `AuthenticatorDevice` object with others data and to do it, we have to keep it like `challenge` between both phases. see [comment](https://github.com/MasterKale/SimpleWebAuthn/issues/95#issuecomment-771062689) also.

For the `metadata`, the value should be returned by `MDS` but if it's not included, we should have at least this `{"authenticator_attachment":"cross-platform"}` or `{"authenticator_attachment":"platform"}` regarding the type of the authenticator registered. This can be very helpful to display dedicated section in the UI regarding the authenticator type.

All this new values should be stored inside `AuthenticatorDevice` object and in the database as well, like this they can be used during verification process.

MasterKale:
I'm glad you brought this up, it encapsulates something I've been musing over lately, one whether I need to think about providing additional date out of attestation verification. I've heard some one or two people that aaguid would be useful output, and now you're bringing up things like authenticator attachment. 

Let me take a look at that developer guide and see if that can't become the basis for returning more info from an attestation verification.

akanass:
In the [guide](https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/WebAuthn_Server_Registration.html), they are saying that:

- If you care about attestation, it is recommended to also store the raw attestation object as part of the credential. This enables you to retroactively inspect credential attestations in response to policy changes and/or compromised authenticators.
- 
- If you care about attestation, use the attestationTrusted, attestationType and attestationMetadata fields to enforce your attestation policy.

akanass:
And [User Handle](https://developers.yubico.com/WebAuthn/WebAuthn_Developer_Guide/User_Handle.html) is very important as well because it can allow usernameless/passwordless verification so we have to store it during `attestation phase` and validate it during `assertion phase`

akanass:
[...]

PR #97, opened 2021-02-02
feature/better-passwordless-usernameless

MasterKale:
To better support WebAuthn's passwordless and usernameless capabilities, additional info should be persisted after a successful attestation verification. This PR achieves this with the following changes:

The return value of **`verifyAttestationResponse()`** has been enhanced with additional information:

- Values are now contained on `attestationInfo`
  - `credentialID` as a `Buffer` (replacing `base64CredentialID`)
  - `credentialPublicKey` as a `Buffer` (replacing `base64PublicKey`)
  - `aaguid` as a UUID `string`
  - `credentialType` as an attestation/assertion's [PublicKeyCredential `[[type]]` slot](https://w3c.github.io/webauthn/#dom-publickeycredential-type-slot) as a `string`
  - `userVerified` as the `boolean` value of `flags.uv`  (now only when the attestation is verified)
  - `attestationObject` as the decoded `response.attestationObject` Buffer returned from the browser

Additionally, **`verifyAssertionResponse()`** has had its return data structure change to align with the above changes. 

- Values are now contained on `assertionInfo`
  - `credentialID` Buffer replaces the old `base64CredentialID` string
  - `newCounter` renames the old `counter`

The `excludeCredentials` argument of **`generateAttestationOptions()`** and `allowCredentials` of **`generateAssertionOptions()`** now expect the credential `id` to be a Buffer. Previously this was expected to be a string. This change was made to make it easier to feed the return value of `verifyAttestationResponse()` directly into future calls to `generateAttestationOptions()` and `generateAssertionOptions()`. This also reduces the amount of encoding/decoding between Buffer and Base64URL throughout attestation and assertion verification, which should be more performant.

Finally **`MetadataService`** picked up some preliminary log statements that had to be commented out at the last second. Enabling this logging will be the target of the next release pending a solution to [anylogger Issue #17](https://github.com/Download/anylogger/issues/17).

## Related Issues
- Issue #95 
- Issue #96


akanass:
@MasterKale I can't see the `UserHandle` generation in your PR. Did you revert it? because you said that it'll be [auto-generated](https://github.com/MasterKale/SimpleWebAuthn/issues/95#issuecomment-772128439) but in the code, we have to pass it in option

And what is the difference between `credentialId` and `aaguid` ?

akanass:
Ok I saw the [comment](https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/server/src/attestation/verifyAttestationResponse.ts#L260) in the code

So if I understant `aaguid` is used for `MetadatService` I guess? and `credentialID` to link the `public-key` with the credential

MasterKale:
Yes, that's correct about `aaguid`. If you want to dive into a weeds a bit [it's a part of the FIDO2 spec](https://fidoalliance.org/specs/fido-v2.0-rd-20180702/fido-metadata-statement-v2.0-rd-20180702.html#authenticator-attestation-guid-aaguid-typedef), and can be passed into MetadataService to retrieve the corresponding [FIDO2 Metadata Statement](https://fidoalliance.org/specs/fido-v2.0-rd-20180702/fido-metadata-statement-v2.0-rd-20180702.html#metadata-statement-format).

`credentialID` and `credentialPublicKey` are a pair of values that must be maintained with each other. You'll pass `credentialID` into `allowCredentials` and `excludeCredentials`, and look up a `credentialPublicKey` by `credentialID` when trying to find the corresponding public key to verify an assertion.

akanass:
@MasterKale and what about `userHandle` generation? we have to do it by ourself? because you told me last time you will auto-generate it

MasterKale:
I made the decision _not_ to generate a `userHandle` because the combination of RP ID and user handle are how authenticators internally track credentials. If the user handle is unique on every attestation (if SimpleWebAuthn were to automatically generate a userHandle every time) then the RP would lose track of credentials and be unable to pull up that credential for use when the user completes an assertion.

This would make it impossible for the RP to support usernameless in particular, as it's the `userHandle` that gets returned during assertion that helps the RP figure out which user is logging in. Therefore the RP needs to be particular about what value is provided here so that it has a chance at correctly associate an attestation credential to a specific user.

akanass:
@MasterKale thanks for your explanation so I will generate by myself by I can see the type is `string` and in documentation it's a `buffer`. What about that?

MasterKale:
`user.id` in the spec's `PublicKeyCredentialCreationOptions` is indeed a `buffer`. However, `generateAttestationOptions()` returns a JSON-compatible representation of options as `PublicKeyCredentialCreationOptionsJSON`. `userID`, then, is defined as a string so that it can be easily transmitted to the front end, at which point Browser's `startAttestation()` turns it into a UInt8Array before passing it to `navigator.credentials.create()`. 

By making `userID` a `string` it leaves it up to the RP to decide how to generate and encode the user handle.

PR #105, opened 2021-02-19
Issue #104-Export-SimpleWebAuthn-types

### LINKED ISSUE #104: Missing exported type `VerifiedAttestation`
The response of [verifyAttestationResponse](https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/server/src/attestation/verifyAttestationResponse.ts#L275) isn't exported in `types` package or `server` package.

TS is going inside `@simplewebauthn/server/dist/attestation/verifyAttestationResponse` to get it.

JayHelton:
This PR fixes #104.

Description:
- exporting the options and response types

Notes:
I initially changed the functions to named imports, instead of default exports, but then changed my mind since these are just type exports

Tested using `yarn link` with the example application. So run `yarn link` in `./packages/server` and then `yarn link "@simplewebauthn/server"` in the root of the example app.

Then in the example app, change the code to look similar to this:

``` 
  let verification: VerifiedAttestation;
  const verificationResponseOptions: VerifyAttestationResponseOpts = {
    credential: body,
    expectedChallenge: `${expectedChallenge}`,
    expectedOrigin,
    expectedRPID: rpID,
  };
  try {
    verification = await verifyAttestationResponse(verificationResponseOptions);
  } catch (error) {
    console.error(error);
    return res.status(400).send({ error: error.message });
  }
  ```
  
  After this is merged, we will be able to import these types like so:
  ```
  import {
  // Registration ("Attestation")
  generateAttestationOptions,
  verifyAttestationResponse,
  // Login ("Assertion")
  generateAssertionOptions,
  verifyAssertionResponse,
  GenerateAssertionOptionsOpts,
  VerifiedAttestation,
} from '@simplewebauthn/server';
```


packages/server/src/attestation/verifyAttestationResponse.ts
```diff
@@ -44,8 +44,8 @@ type Options = {
  * @param supportedAlgorithmIDs Array of numeric COSE algorithm identifiers supported for
  * attestation by this RP. See https://www.iana.org/assignments/cose/cose.xhtml#algorithms
  */
-export default async function verifyAttestationResponse(
-  options: Options,
+export async function verifyAttestationResponse(
```
JayHelton, line 47: Let me know if youd rather keep this as the default export.

I just did what i would normally do if a module went from a single public export, to multiple exports

JayHelton:
derp, types dont export the same as interfaces

packages/server/src/assertion/generateAssertionOptions.ts
```diff
@@ -8,7 +8,7 @@ import base64url from 'base64url';
 
 import generateChallenge from '../helpers/generateChallenge';
 
-type Options = {
+export type GenerateAssertionOptions = {
```
MasterKale, line 11: I think this and `GenerateAttestationOptions` types should have an additional `Options` added to the end because they're "options for the `generateAttestationOptions` and `generateAssertionOptions` methods":

Example:
- `GenerateAssertionOptions` -> `GenerateAssertionOptionsOptions`

Perhaps use `Opts` instead? `GenerateAssertionOptionsOpts`? 🤔

packages/server/src/assertion/verifyAssertionResponse.ts
```diff
@@ -12,7 +12,7 @@ import verifySignature from '../helpers/verifySignature';
 import parseAuthenticatorData from '../helpers/parseAuthenticatorData';
 import isBase64URLString from '../helpers/isBase64URLString';
 
-type Options = {
+export type VerifyAssertionOptions = {
```
MasterKale, line 15: I'd like these types to follow the "method name + Options" naming convention. That'd make this `VerifyAssertionResponseOptions`

packages/server/src/attestation/generateAttestationOptions.ts
```diff
@@ -11,7 +11,7 @@ import base64url from 'base64url';
 
 import generateChallenge from '../helpers/generateChallenge';
 
-type Options = {
+export type GenerateAttestationOptions = {
```
MasterKale, line 14: Same note as with `GenerateAssertionOptions` above, please tack on an extra `Options`.

packages/server/src/attestation/verifyAttestationResponse.ts
```diff
@@ -20,7 +20,7 @@ import verifyTPM from './verifications/tpm/verifyTPM';
 import verifyAndroidKey from './verifications/verifyAndroidKey';
 import verifyApple from './verifications/verifyApple';
 
-type Options = {
+export type VerifyAttestationOptions = {
```
MasterKale, line 23: And same note here as with the `VerifyAssertionOptions`, this should be `VerifyAssertionResponseOptions` instead

packages/server/src/index.ts
```diff
@@ -2,16 +2,22 @@
...
+import verifyAttestationResponse, { VerifiedAttestation, VerifyAttestationOptions } from './attestation/verifyAttestationResponse';
+import verifyAssertionResponse, { VerifiedAssertion, VerifyAssertionOptions } from './assertion/verifyAssertionResponse';
 import MetadataService from './metadata/metadataService';
 
 export {
   generateAttestationOptions,
   verifyAttestationResponse,
   generateAssertionOptions,
   verifyAssertionResponse,
+  GenerateAttestationOptions,
```
MasterKale, line 16: Can we use `export type` here with these instead? If so we should probably `import type` the types above instead.

MasterKale:
After a bit more thought I think you should use `Opts` instead of an additional `Options` in my comments above:

Ex:
- `GenerateAssertionOptions` -> `GenerateAssertionOptionsOpts`
- `VerifyAssertionOptions` -> `VerifyAssertionResponseOpts`

packages/server/src/index.ts
```diff
@@ -2,16 +2,22 @@
  * @packageDocumentation
  * @module @simplewebauthn/server
  */
-import generateAttestationOptions from './attestation/generateAttestationOptions';
-import verifyAttestationResponse from './attestation/verifyAttestationResponse';
-import generateAssertionOptions from './assertion/generateAssertionOptions';
-import verifyAssertionResponse from './assertion/verifyAssertionResponse';
+import generateAttestationOptions, { GenerateAttestationOptions } from './attestation/generateAttestationOptions';
+import generateAssertionOptions, { GenerateAssertionOptions } from './assertion/generateAssertionOptions';
+import verifyAttestationResponse, { VerifiedAttestation, VerifyAttestationOptions } from './attestation/verifyAttestationResponse';
```
MasterKale, line 7: Can you please swap this line with the line above? It should make for a cleaner diff to make it look more like the diff for Line 5.

packages/server/src/assertion/verifyAssertionResponse.ts
JayHelton: Noted!

MasterKale (approved):
lgtm :shipit:

MasterKale:
These changes have been published in **@simplewebauthn/server@2.2.0**

PR #106, opened 2021-02-22
fix: public key conversion

chaotixkilla:
This PR fixes an issue occuring when validating assertion. When the process enters `convertPublicKeyToPEM()`, the following error is thrown:

```
TypeError: Cannot read property 'get' of undefined
  at Object.convertPublicKeyToPEM [as default] (/home/<user>/Desktop/<my_project>/node_modules/@simplewebauthn/server/dist/helpers/convertPublicKeyToPEM.js:18:24)
  at Object.verifyAssertionResponse (/home/<user>/Desktop/<my_project>/node_modules/@simplewebauthn/server/dist/assertion/verifyAssertionResponse.js:128:54)
```

I believe this is caused because commit `548aa64bbbcfc70491a7ce2b103f9bd78ac010b6` changes the function to accept `Buffer`, instead of `string`, but the argument `publicKey` is being sent as a Buffer object and not a Buffer, resulting in the error above

### DIFF
--- a/packages/server/src/helpers/convertPublicKeyToPEM.ts
+++ b/packages/server/src/helpers/convertPublicKeyToPEM.ts
@@ -6,7 +6,8 @@ import { COSEKEYS, COSEKTY, COSECRV } from './convertCOSEtoPKCS';
 export default function convertPublicKeyToPEM(publicKey: Buffer): string {
   let struct;
   try {
-    struct = cbor.decodeAllSync(publicKey)[0];
+    const pkBuffer = Buffer.from(publicKey);
+    struct = cbor.decodeAllSync(pkBuffer)[0];
   } catch (err) {
     throw new Error(`Error decoding public key while converting to PEM: ${err.message}`);
   }


akanass:
You have to pass the buffer by yourself before instead of doing it inside the function. I am doing like this in my project and all is working well

chaotixkilla:
@akanass Either way, shouldn't the example project rectify this issue then? I'm basing my project off the example one, and the conversion isn't mentioned anywhere from what I've seen

MasterKale:
I apologize for the confusion. The example project doesn't do any such conversion because the `AuthenticatorDevice` type was updated to make `credentialPublicKey` into a Buffer, and so the example project now tracks all of these as `Buffers` within its `inMemoryUserDeviceDB`:

https://github.com/MasterKale/SimpleWebAuthn/blob/31ba87cb189b723b053a1e1edb12b173200a5b39/packages/typescript-types/src/index.ts#L121

https://github.com/MasterKale/SimpleWebAuthn/blob/31ba87cb189b723b053a1e1edb12b173200a5b39/example/example-server.d.ts#L37

(the `LoggedInUser` type defined in the example project references the `AuthenticatorDevice` type)

`verifyAssertion()`, then, expects the `credentialPublicKey` to be a buffer as a property on the `AuthenticatorDevice` you pass in as the `credential` argument:

https://github.com/MasterKale/SimpleWebAuthn/blob/31ba87cb189b723b053a1e1edb12b173200a5b39/packages/server/src/assertion/verifyAssertionResponse.ts#L20

PR #115, opened 2021-03-31
fix/issue-111-cert-hExtV-parse-error

MasterKale:
This issue adds in a workaround to certificate info parsing to support verification of Packed attestation responses from Chrome WebAuthn dev tool virtual authenticators.

## Context:

Chrome's WebAuthn dev tool enables developers to use "virtual" software authenticators in place of typical authenticator hardware. This PR is intended to ensure that SimpleWebAuthn can handle leaf certs, such as the ones in these virtual authenticators, that specify the byte sequence [`\x30\x03\x01\x01\x00`](https://source.chromium.org/chromium/chromium/src/+/master:device/fido/virtual_fido_device.cc;drc=184bf7b03160ebf2217ab4711118d013a750908a;bpv=1;bpt=1;l=552?gsn=kBasicContraintsContents&gs=kythe%3A%2F%2Fchromium.googlesource.com%2Fchromium%2Fsrc%3Flang%3Dc%252B%252B%3Fpath%3Dsrc%2Fdevice%2Ffido%2Fvirtual_fido_device.cc%23IOOK5Gp8xPpETdIy6gegBWULXOp1WINqit2QnWGs1Fg&gs=kythe%3A%2F%2Fchromium.googlesource.com%2Fchromium%2Fsrc%3Flang%3Dc%252B%252B%3Fpath%3Dsrc%2Fdevice%2Ffido%2Fvirtual_fido_device.cc%23YnpuSh_lp-fg0kDnhqWe1BnH79fOiNaYU6DBqZLAZD8) for the cert's Basic Constraints extension.

As of March 2021 the jsrsasign@^10.0.5 library has a hardcoded check for `"30030101ff"`, but not `"3003010100"` (notice the difference between `"ff"` and `"00"` at the end), indicating whether or not this is a certificate authority certificate:

https://github.com/kjur/jsrsasign/blob/482e651f2bb380dad3da4bbf0ae220fe3021d407/src/x509-1.1.js#L660

Physical hardware authenticators have been observed to specify `"3000"` for this constraint; this value evaluates to `!!undefined` => `false`, satisfying the Packed attestation verification's requirement that, "the Basic Constraints extension MUST have the CA component set to false":

https://w3c.github.io/webauthn/#sctn-packed-attestation-cert-requirements

**@simplewebauthn/server** will have to implement its own workaround until this issue is resolved in jsrsasign.


MasterKale:
After sleeping on it I awoke to the realization that this PR is laughably complex. What I'm going to do instead is try and lean more on the `asn1-schema` library to peel apart X.509 certs into a sensible data structure.

MasterKale:
The issue this fixes was a bug in Chrome's virtual authenticators:

https://chromium-review.googlesource.com/c/chromium/src/+/2797998/

I'll leave this open for now because I still think it might be a good idea to decrease the project's dependency on jsrsasign; it's kinda unwieldy to use and if there's a chance to remove a dependency from the project I'm willing to entertain the notion.

MasterKale:
A few days later I think it's best to wrap this PR up with the slight refactor to `getCertificateInfo()` to use `@peculiar/asn1-schema` instead. `jsrsasign` does a tremendous job of handling signature verification of various kinds, so after a bit of experimentation I've decided there's no need to remove it completely.

Issue #114, opened 2021-03-30
What to do about `supportsWebAuthn()`...


MasterKale:
As highlighted in PR #112 and #113, **@simplewebauthn/browser**'s `supportsWebAuthn()` method can't actually execute in browsers that _don't_ support WebAuthn, like IE11 and Edge Legacy, because those browsers don't support ES2018 syntax. This negates the reason for its existence.

As I mentioned in PR #113, the path forward is ambiguous to me. It's likely not as simple as updating **packages/browser/tsconfig.json** to target ES5 instead of ES2018.

Possible options I proposed included:
- Maybe there's a way to build the code with Webpack that lets developers pick-and-choose which version of the library they want to use in their code.
- Maybe the solution is another library that's only the `supportsWebAuthn()` method built for ES5 environments.
- A third alternative is I just nuke `supportsWebAuthn()` and update the docs to say, "if you need to support IE11/Edge Legacy, here's a snippet of VanillaJS you can copy-paste into your code to tell if WebAuthn is available."

I'm open to suggestions on how we might solve the problem of building a library that simultaneously needs to _sometimes_ support ES5 while _often_ supporting ES2018.

MasterKale:
For the record I'm leaning towards option 3️⃣ because it's the easiest to pull off 😂

akanass:
@MasterKale The best solution is to stop using `Webpack` because this tool is old school and to have a better build system with multiple targets, the right tool to use is `rollup`.

I use it in my projects and even with the one that supports this library, I have no compilation problems and I output files in MJS for browsers supporting imports of ESM modules or even SystemJS compatible files with a import modules inside.

akanass:
But once again, why compile this library in ES5 when the browsers that support WebAuthn all support ES2018 compilation.

Older browsers don't support WebAuthn so I don't see the point of all of this.

mariusmarais:
@akanass I assume the underlying issue is that this non-ES5 module is blowing up the entire app. For the app WebAuthn is optional and only available in newer browsers. But the pure existence of the option is breaking the app in the older browsers. Since the app must work in older browsers because of reasons, the WebAuthn functionality not cannot be used by anyone. (This is putting a lot of words in @Moumouls 's mouth, I hope the gist is correct.)

[...]

PR #117, opened 2021-04-05
fix/issue-114-browser-multi-bundle

MasterKale:
This PR updates @simplewebauthn/browser's build and bundling pipeline to generate artifacts to support a wider range of use cases. This includes:

- A new `"main"` build that targets ES2018
- A new `"browser"` build that targets ES5
- A UMD bundle that targets ES2018
- A UMD bundle that targets ES5 for deprecated browsers

This should address Issue #114.


akanass:
I have done all optimisations inside my [PR](https://github.com/MasterKale/SimpleWebAuthn/pull/116) and all is much better now

MasterKale:
A note to myself for tomorrow: why not update @swan/browser to not use as modern JS? If I refactor it to use "older"/more explicit JS I think there's a chance to avoid needing to fuss with tslib and multiple UMD bundles. The structure of the WebAuthn API is largely set for now, no need to "optimize" for unpacking future values via the spread operator (that's the real reason for jumping through all these bundling hoops)

MasterKale:
I fired up BrowserStack and got some screenshots of the new ES5 UMD bundles in action, specifically testing support for `supportsWebAuthn()` after linking to the bundle in a `<script>` tag:

```html
<!-- Manually renamed dist/es5/index.umd.min.js -->
<script src="../index.umd.es2018.min.js"></script>
<script>
  (function() {
    const elemBegin = document.getElementById('btnBegin');
    const elemSuccess = document.getElementById('success');
    const elemError = document.getElementById('error');
    
    const supportsWebauthn = SimpleWebAuthnBrowser.supportsWebauthn;
    
    // Hide the Begin button if the browser is incapable of using WebAuthn
    if (!supportsWebauthn()) {
      elemBegin.style.display = 'none';
      elemError.innerText = "It seems this browser doesn't support WebAuthn...";
      return;
    }
    
    // ...
  })();
</script>
```

### IE 11
swan-browser-ie11-es5-umd

### Edge 17
swan-browser-edge17-es5-umd

MasterKale:
I'm 99% satisfied this is ready to go.

That last 1% will involve me sleeping on this and then finding time in the next day or two to spin up a fresh Create-React-App project, [attempt to polyfill enough of it to run in IE11](https://github.com/facebook/create-react-app/tree/master/packages/react-app-polyfill), and then try to use @simplewebauthn/browser as I would when building a typical SPA. If I can get `supportsWebAuthn()` to work in IE11 and Edge 17 then I'll publish this.

akanass:
The job seems to have been done well, happy to see it happen successfully

MasterKale:
Here's `supportsWebAuthn()` working in IE10 in a production build of a Create-React-App project:

Screen Shot 2021-04-08 at 8 31 52 AM

I only needed to do a couple of things with the React project:

1. `npm install react-app-polyfill`
2. Add `import 'react-app-polyfill/ie9';` to the top of **src/index.js**

And when I flipped over to TypeScript, I just needed to make sure **tsconfig.json** was set up with `"target": "es5"` and everything worked just fine.

akanass:
@MasterKale so all was done successfully :D I am glad to hear that

PR #119, opened 2021-04-15
fix/issue-114-es5-better-build-version-and-as-default

akanass:
This PR updates @simplewebauthn/browser's build and bundling pipeline to generate artifacts to support a wider range of use cases. This includes:

A new `"main"` build that targets ES5
A new `"main:es2018"` build that targets ES2018
A UMD bundle that targets ES2018
A UMD bundle that targets ES5 for deprecated browsers

`ES5` version has been optimised to have a better weight at the end and compatible with `browser` and `node` build systems.

The documentation has been updated as well to explain `ES5` is now the default version and how to use `ES2018` in build systems.

This should address Issue #114 and especially for framework with `SSR` compatibility.

You can see the result of the new build with `Next.js` [here](https://github.com/MasterKale/SimpleWebAuthn/issues/114#issuecomment-820461287)


akanass:
@MasterKale this PR makes a better `ES5` version so we have to use it anyway.

If you want to have `ES5` as default you can use all of this else I have juste to rollback `package.json` and `README` and `ES2018` will still be the default version but the new `ES5` has to be used because it's an optimised one.

akanass:
As discussed with @Moumouls in #114 we agreed that `ES5` has to be the default one to support build and rendering in client and server side without any trick like this more people will be able to use it in old and modern browsers.

packages/browser/README.md
```diff
@@ -28,26 +28,28 @@ npm install @simplewebauthn/browser
 
 ### UMD
 
-This package can also be installed via **unpkg** by including the following script in your page's `<head>` element. The library's methods will be available on the global **`SimpleWebAuthnBrowser`** object.
+This package can also be installed via **unpkg** by including the following script in your page's `<head>` element. The
+library's methods will be available on the global **`SimpleWebAuthnBrowser`** object.
```
MasterKale, line 32: Please remove these "line-break after N characters".

packages/browser/README.md
```diff
@@ -28,26 +28,28 @@ npm install @simplewebauthn/browser
...
+#### ES5
 
-If you only need to support modern browsers, include the `ES2018` version:
+If you need to support WebAuthn feature detection in deprecated browsers like IE11 and Edge Legacy, include the `ES5`
+version:
 
 ```html
 
-<script src="https://unpkg.com/@simplewebauthn/browser/dist/es2018/index.umd.min.js"></script>
+<script src='https://unpkg.com/@simplewebauthn/browser/dist/es5/index.umd.min.js'></script>
```
MasterKale, line 43: Something replaced double-quotes with single-quotes, can you revert those for the sake of a cleaner PR?

packages/browser/package.json
```diff
@@ -2,8 +2,8 @@
   "name": "@simplewebauthn/browser",
   "version": "2.2.1",
   "description": "SimpleWebAuthn for Browsers",
-  "main": "dist/es2018/index.js",
-  "browser": "dist/es5/index.js",
+  "main": "dist/es5/index.js",
+  "jsnext:main": "dist/es2018/index.js",
```
MasterKale, line 6: I think a better name for this could be `"main:es2018"`.

akanass: `jsnext:main` is used in a lot of library to reference this kind of build but if you want to have `main:es2018` I'll adapt it.

packages/browser/README.md
akanass: fixed in next commit

packages/browser/README.md
akanass: fixed in next commit

packages/browser/README.md
```diff
@@ -7,16 +7,16 @@
...
-    - [UMD](#umd)
-      - [ES2018](#es2018)
-      - [ES5](#es5)
-  - [Usage](#usage)
-  - [Building for Production](#building-for-production)
-    - [ES2018](#es2018-1)
-    - [ES5](#es5-1)
-      - [Webpack support](#webpack-support)
-      - [Rollup support](#rollup-support)
+    - [Installation](#installation)
```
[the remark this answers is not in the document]
akanass: fixed in next commit

packages/browser/package.json
akanass: fixed in next commit

akanass:
@MasterKale the requested changes have been made. You'll have to decide how to bump the version because we don't have the same default version anymore so I think `2.3.0` would be the minimum

packages/browser/package.json
MasterKale: Huh, looking into how common the use of `"jsnext:main"` is, I'm seeing more recent conversations highlighting that `"module"` is a more standardized version of this property (mostly from Rich Harris in Rollup-related issues like this one: https://github.com/rollup/rollup/wiki/pkg.module).

Should this be `"module"` instead of even something like `"main:es2018"`?

MasterKale: Even Webpack uses `"module"` in its `resolve.mainFields`: https://webpack.js.org/configuration/resolve/#resolvemainfields

akanass: @MasterKale we can't use `module` because this field is before `main` in the hierarchy of fields resolution.

So if we're using `module` it'll be the `ES2018` version in the build process and not `ES5` that's why I'm using a custom field to be sure runtime like `Next.js` in `SSR` will use the right version.

I know all about `browser`, `module` and `main` fields and we can only use them if the system behind will use the correct field during build process.

Normally, `main` is for `commonjs` version then `module` is for `esm` version and `browser` for dedicated version.

But in our case, we've two versions in `esm` and we want the `main` in `ES5` so we can't use the `module` field for all what I said before.

What we've now it's the right way to do, believe me :)

MasterKale:
I just tested this out in Next.JS and Create-React-App in IE10 and IE11 and everything seems fine. Thank you for another speedy build improvement! 😃

akanass:
@MasterKale It was my pleasure again to help on it and glad you're happy with this 😄

PR #130, opened 2021-06-17
feat/default-to-es2018-bundle

MasterKale:
This PR reorganizes @simplewebauthn/browser to use the ES2018 bundle for `"main"` in **package.json**. The ES5 CommonJS bundle will no longer be maintained. Both the ES2018 and ES5 UMD bundles will continue to be maintained for now.

I've also cleaned up the README to remove references to the CJS ES5 bundle.


MasterKale:
Nevermind, #129 has some other good stuff in it like dep upgrades, better unpkg support, and a simpler file structure.

PR #138, opened 2021-08-02
feat/settings-service

MasterKale:
This PR adds a new `SettingsService` singleton for managing various configurable aspects of @simplewebauthn/server. The first such configurable options are the root certs used for verifying registration responses containing `android-safetynet` and `apple` attestation statement formats.

The service allows for setting root certificates as either PEM-formatted `string`'s, or `Buffer`'s:

```ts
import { SettingsService } from '@simplewebauthn/server';

SettingsService.setRootCertificate({
  attestationFormat: 'android-safetynet',
  // Buffer
  certificates: [
    fs.readFileSync(path.resolve(__dirname, './GSRCA.crt')),
    fs.readFileSync(path.resolve(__dirname, './GSR2.crt')),
  ],
});

SettingsService.setRootCertificate({
  attestationFormat: 'apple',
  // PEM-formatted String
  certificates: [
    fs.readFileSync(path.resolve(__dirname, './Apple_WebAuthn_Root_CA.pem'), { encoding: 'utf-8' }),
  ],
});
```

Certificates for a format can be retrieved with the `fmt` of the attestation statement:

```ts
// Array of PEM-formatted strings (or an empty array if no cert is set for the given format)
const rootCertificates: string[] = settingsService.getRootCertificate({ attestationFormat: fmt });
```

When root certificates are registered for an attestation format, an attempt will be made to validate the certificate path using each root cert until a complete path is formed. **If an empty array is set for a specific attestation format then certificate path validation will not occur!**

The following root certificates are now registered via this API as default certificates for the following formats:

- `"android-key"` ([the two root certs specified here](https://developer.android.com/training/articles/security-key-attestation#root_certificate))
- `"android-safetynet"` (GlobalSign Root CA, GlobalSign Root 2)
- `"apple"` (Apple WebAuthn Root CA)

Root certs can be specified for all attestation formats. The specific certificates mentioned above can be overwritten by calling `SettingsService.setRootCertificate()` with the same `attestationFormat` identifier and a new array of certificates.

### A quick note...

**How** these certificates are read is intentionally left up to the RP developer. There are a lot of different ways to retrieve these certificates, including HTTPS requests out to the internet, network requests to vaults on an intranet, direct access to the filesystem...trying to define a method to support all such potential use cases would quickly turn into a maintenance nightmare. Use of `SettingsService` is a more advanced feature of the library that most developers can ignore, so I have chosen to leave it to developers who opt into its use to retrieve certificates in a way that's most appropriate for their implementation.


MasterKale:
@jstewmon If this comment makes it your way I'd love your feedback on this feature. It was inspired by our conversation in #125.

packages/server/src/services/settingsService.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Google GlobalSign R2
+ *
+ * Downloaded from https://pki.goog/gsr2/GSR2.crt
+ *
+ * Valid until 2021-12-15 @ 00:00 PST
+ */
+settingsService.setRootCertificate({
+  attestationFormat: 'android-safetynet',
+  certificate: fs.readFileSync(path.resolve(__dirname, './defaultRootCerts/GSR2.crt')),
```
jstewmon, line 56: I didn't see any packaging updates to facilitate this file loading strategy. tsc will only produce .js and .d.ts output in the dist directory, so you'll need to copy these into the desired location with another tool. Or, you could place them elsewhere (e.g. `server/ca/GSR2.crt`), add `ca/*.crt` to the `files` array in `package.json` and resolve the path using [pkg-dir] or similar.

[pkg-dir]: https://www.npmjs.com/package/pkg-dir

packages/server/src/services/settingsService.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Allow setting custom root certificates for attestation formats that use them
+   *
+   * The certificate can be specified as a raw `Buffer`, or as a PEM-formatted string. If a
+   * `Buffer` is passed in it will be converted to PEM format.
+   */
+  setRootCertificate(opts: {
+    attestationFormat: AttestationFormat;
+    certificate: Buffer | string;
```
jstewmon, line 23: Would this support multiple certificates (i.e. can multiple pem certs be in the string)? I know `verifyAndroidKey` isn't currently verifying the root ca, but I think some folks will want to only allow attestation statements that verify the key is in hardware-secure storage, for which there are [two keys]. Multiple root ca certificates may be needed to handle transitions in the future.

[two keys]: https://developer.android.com/training/articles/security-key-attestation#root_certificate

packages/server/src/services/settingsService.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    this.pemCertificates.set(attestationFormat, newCertificate);
+  }
+
+  /**
+   * Get any registered root certificates for the specified attestation format
+   */
+  getRootCertificate(opts: { attestationFormat: AttestationFormat }): string {
+    const { attestationFormat } = opts;
+    return this.pemCertificates.get(attestationFormat) ?? '';
```
jstewmon, line 40: Why return `''` as a sentinel value?

I'm not sure what happens if an empty string is pushed onto the certificate chain in the verify functions, but my guess is that it will fail with an error that won't clearly point to the problem being that a root certificate wasn't registered.

My initial thought was that this should throw an error if `!this.pemCertificates.has(attestationFormat)`, but I'm not sure if there is some valid case for requesting a certificate which has not been registered. If there is, then explicitly registering `null` and checking for `null` in the verify functions might be more robust.

jstewmon (commented):
👋 Hi @MasterKale, sorry for going dark on #125 - on top of regular priority juggling, I changed jobs, and I'm no longer actively working with the library.

Anyway, I had a couple of comments for you, aside from the limitations we previously discussed about using a singleton. (I'm not providing any feedback on the singleton strategy here, since we discussed those on #125.)

packages/server/src/services/settingsService.ts
MasterKale: Doh, you're right, I need to update the build to include these in **dist/** 🤦‍♂️ 

`pkg-dir` is a new to me, I'll take a look at that.

packages/server/src/services/settingsService.ts
MasterKale: Not the way it's currently implemented, no. I suppose it'd be best to make this all "certificate**s**" with `Buffer[] | string[]` and then try a verification for each cert passed in...but then that'd only allow for a single root cert per execution. What about cert chains? `Buffer[][] | string[][]` and then `forEach` with an inner `concat`?

packages/server/src/services/settingsService.ts
MasterKale: I went this route for cleaner typing, if the method is guaranteed to return a string then that's one less type guard I'd need to write 😅 

My gut tells me it's better to just return whatever `.get()` returns, which would make this `string | undefined`. Then the verification method can be updated to check the value of `rootCertificate` and impose its restrictions on what should have been returned.

> ...my guess is that it will fail with an error that won't clearly point to the problem being that a root certificate wasn't registered.

The cert path validation code would likely error out that the value provided wasn't a valid certificate structure. I kinda think if you're using this feature you'll be smart enough to know how to properly specify a PEM cert, since this is a Power User feature.

packages/server/src/services/settingsService.ts
MasterKale: I've updated this all to support an array of multiple root certs, as buffers and/or strings.

packages/server/src/services/settingsService.ts
MasterKale: Since tweaking SettingsService to support setting multiple possible root certs this method has been updated to return an empty array instead. It makes sense to me to do so, since the logic later on that uses these values checks if `rootCertificates.length === 0` and branches appropriately.

packages/server/src/services/settingsService.ts
MasterKale: I punted on this a bit by saving these certs in PEM format as .ts files and importing them like any other string.

Issue #160, opened 2021-09-20
Idea: Add new helper methods for MFA, PWL, and Usernameless registration and authentication


MasterKale:
SimpleWebAuthn already makes interacting with the WebAuthn library fairly painless, but can I take things one step further? The thought occurred to me that I might be able to offer a higher-level abstraction to the registration and authentication options generation and response verification methods. These abstractions would set combinations of options appropriate for use of WebAuthn for MFA, Passwordless, and Usernameless "modes", with minimal arguments to specify things like `allowCredentials`:

**Registration**

- `getMFARegistrationOptions()`
- `getPWLRegistrationOptions()`
- `getUNLRegistrationOptions()` (UNL -> "Usernameless", you heard it here first! 😅)

- `verifyMFARegistrationResponse()`
- `verifyPWLRegistrationResponse()`
- `verifyUNLRegistrationResponse()`

**Authentication**

- `getMFAAuthenticationOptions()`
- `getPWLAuthenticationOptions()`
- `getUNLAuthenticationOptions()`

- `verifyMFAAuthenticationResponse()`
- `verifyPWLAuthenticationResponse()`
- `verifyUNLAuthenticationResponse()`

Or maybe a new, optional `mode` argument on the existing methods that you specify as `"mfa" | "pwl" | "unl"`? That might actually be cleaner 🤔 

In any case the idea is, there are certain options you _must_ set if you want to, for example, leverage WebAuthn for Passwordless login. Without digging into the docs, though, you might not think to set them. I think there's a chance here to help devs even further by offering a quick and easy way to generate appropriate options and verify responses that will help ensure secure authentication for all use cases.

JayHelton:
Before you ask, yes im subscriabed to the repo when new issues are update. 😆 

I like the idea of having abstractions between resident keys and non resident keys using a terminology that fits the UX. I am not confident in that abstraction being MFA / PWDL, or UNL.  What is the different between PWDL versus MFA in terms of the webauthn impl details? Off the top of my head, I cant think of any. That might change my opinion though.

In general, i prefer method abstractions over parametered abstractions, like the mode argument proposal. Then again, Im biased towards a "functional" approach over imperitive.

MaKleSoft:
[...]

Issue #140, opened 2021-08-18
Export helpers for use outside of the library


MasterKale:
Some advanced use cases would benefit from access to the various helpers within **@simplewebauthn/server**. I plan on exporting all of the helpers in one large `helpers` export that will expose logic to those who are, for example, looking to build something that supports attestation via both WebAuthn and Apple's App Attest.

See https://github.com/MasterKale/SimpleWebAuthn/discussions/124 for some more background context.

MasterKale:
This issue has been resolved in **v4.0.0**! Please see the [changelog](https://github.com/MasterKale/SimpleWebAuthn/releases/tag/v4.0.0) and take a peek at all of the methods available for import from `@simplewebauthn/server/helpers` (not a new package, just a subpath.)

PR #171, opened 2022-01-27
feat/modes

MasterKale:
I started thinking about #160 again, about how I might offer opinionated ways of leveraging WebAuthn. Part of this was recently motivated by my recent realization that "we" (in the "WebAuthn community" sense) need to switch from "WebAuthn is great, have you heard of it" to "WebAuthn is great, here are the small number of ways that you can properly use it".

For a while I didn't feel like I had a good idea for avoiding ugly method names like `getMFARegistrationOptions()` or `verifyPWLAuthenticationResponse()`. Earlier this week, though, I had an epiphany that I can use the same mechanism I used in #140 to export "mode" helpers instead.

## Example

Change this:

```ts
import {
  generateRegistrationOptions,
  verifyRegistrationResponse,
} from '@simplewebauthn/server';
```

To this:

```ts
import {
  generateRegistrationOptions,
  verifyRegistrationResponse,
} from '@simplewebauthn/server/passwordless';
```

And now no matter the options you pass into these methods some options will be overriden to _always_ require user verification so that users fulfill "something you have" and "something you are"/"something you know" in a single authenticator interaction.

Later, if you decide you want lean into discoverable credentials more, you can update the package name and start generating options and verifying responses without needing to know the exact combination of options that will prompt the user to choose a credential to auth with without them needing to enter a username or email address:

```ts
import {
  generateRegistrationOptions,
  verifyRegistrationResponse,
} from '@simplewebauthn/server/passkeys';
```

By keeping the call signatures and return types the same, these opinionated helpers can be swapped out without needing to rewrite any of the code that call the methods.


MasterKale:
I'm abandoning this work because I think it's better left up to integration libraries to control how SimpleWebAuthn gets called. I'll instead focus on updating the docs website to offer guidance on how best to call WebAuthn for those discrete use cases that the community has settled on:

- MFA
- Passwordless
- Passkeys

PR #191, opened 2022-04-03
Change error name to reflect the type

### LINKED ISSUE #190: Differentiate browser / server errors by name
Hi, since v5, looks like the `error.name` is always `WebAuthnError`, so if I want to show a translated message on my app, the only way to distinguish between errors is to check the message (which may change).

Previously this code (which is taken from the examples)
```js
try {
...
} catch(error) {
  if (error.name === 'InvalidStateError') {
    throw ['WEB_AUTHN.ERROR.DEVICE_ALREADY_EXIST'];
  } else if (error.name === 'NotAllowedError') {
    throw ['WEB_AUTHN.ERROR.NOT_ALLOWED'];
  }
}
```

Do you think that it is better to change the name of the error?
I can prepare a quick PR for it.

felixmosh:
fixes #190

### DIFF
--- a/packages/browser/src/helpers/identifyAuthenticationError.ts
+++ b/packages/browser/src/helpers/identifyAuthenticationError.ts
@@ -20,38 +20,42 @@ export function identifyAuthenticationError({
   if (error.name === 'AbortError') {
     if (options.signal === new AbortController().signal) {
       // https://www.w3.org/TR/webauthn-2/#sctn-createCredential (Step 16)
-      return new WebAuthnError('Authentication ceremony was sent an abort signal (AbortError)');
+      return new WebAuthnError('Authentication ceremony was sent an abort signal', 'AbortError');
     }
   } else if (error.name === 'NotAllowedError') {
     if (publicKey.allowCredentials?.length) {
       // https://www.w3.org/TR/webauthn-2/#sctn-discover-from-external-source (Step 17)
       // https://www.w3.org/TR/webauthn-2/#sctn-op-get-assertion (Step 6)
       return new WebAuthnError(
-        'No available authenticator recognized any of the allowed credentials (NotAllowedError)',
+        'No available authenticator recognized any of the allowed credentials',
+        'NotAllowedError',
       );
     }
 
     // https://www.w3.org/TR/webauthn-2/#sctn-discover-from-external-source (Step 18)
     // https://www.w3.org/TR/webauthn-2/#sctn-op-get-assertion (Step 7)
     return new WebAuthnError(
-      'User clicked cancel, or the authentication ceremony timed out (NotAllowedError)',
+      'User clicked cancel, or the authentication ceremony timed out',
+      'NotAllowedError',
     );
   } else if (error.name === 'SecurityError') {
     const effectiveDomain = window.location.hostname;
     if (!isValidDomain(effectiveDomain)) {
       // https://www.w3.org/TR/webauthn-2/#sctn-discover-from-external-source (Step 5)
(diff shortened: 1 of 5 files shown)


MasterKale:
Hmm, I don't like the idea of calling an error "WebAuthnError" and then allowing for its name to be overridden so `error.name` would never be `"WebAuthnError"`. The better way to do this is probably to throw basic `Error`'s instead and set the name to the original WebAuthn error so that RPs like you can continue to rely on `error.name` for detection. Then just drop `"WebAuthnError"` completely so what the functionality really becomes is an overriding of the description of the error.

MasterKale:
I experimented with some potential solutions tonight and after hitting a bunch of dead ends I do think your proposed solution is the only way forward. I'm going to request a fix, and once that's done this should be good to merge.

packages/browser/src/helpers/identifyAuthenticationError.ts
```diff
@@ -20,38 +20,48 @@ export function identifyAuthenticationError({
   if (error.name === 'AbortError') {
     if (options.signal === new AbortController().signal) {
       // https://www.w3.org/TR/webauthn-2/#sctn-createCredential (Step 16)
-      return new WebAuthnError('Authentication ceremony was sent an abort signal (AbortError)');
+      return new WebAuthnError(
+        'Authentication ceremony was sent an abort signal (AbortError)',
```
MasterKale, line 24: Since the error name will now convey this, can you please remove from all error messages the error name in the parenthesis? It's being made redundant by these changes.

e.g. drop "(AbortError)" from "Authentication ceremony was sent an abort signal (AbortError)"

packages/browser/src/helpers/identifyAuthenticationError.ts
```diff
@@ -20,38 +20,48 @@ export function identifyAuthenticationError({
...
+        'AbortError',
```
MasterKale, line 25: Can you also update tests to ensure that the `name` property on all thrown errors is the appropriate value? That seems to me to be the most important thing to assert in all this, that SimpleWebAuthn doesn't nuke the actual error name again and break existing code that relies on it.

MasterKale (changes requested):

packages/browser/src/helpers/identifyAuthenticationError.ts
felixmosh: I've updated the messages, but for some reason I can't run tests locally, fails on typescript :|

MasterKale: I'll bet I know what you're running into. The tests for this functionality are in a sorry state, I was fighting them myself just last night. The issue was I assumed `NotAllowedError`/`SecurityError`/etc... were actual subclasses of `Error` that should exist on `global`. Instead they're simply instances of `Error` with custom `name` properties (because of course they are 🤦) and I wasted a bunch of time trying to mock them out.

The solution will probably be to update the tests to check `name` instead of the type of the error, and then rip out all of the custom error subclasses I defined for the tests. I'll take some time tonight to do this and get it on master so that `test` and `test:watch` resume working as expected. Afterwards you should be able to rebase on them and get tests running locally.

Apologies for the inconvenience 😅

MasterKale: @felixmosh I just merged #194 that should fix the local browser testing issues. Can you rebase off of **master** and try again?

felixmosh:
I've updated tests, thank you

MasterKale (approved):

MasterKale:
@felixmosh Thank you for your contribution! This is a great improvement to the error detection logic 🚀

PR #192, opened 2022-04-10
delete unused import type

TaigaMikami:
Removed imports of types not used in verifyAuthenticationResponse.ts.


MasterKale:
Argh, I should have caught this when I was working on #181. Thank you @TaigaMikami for helping keep things clean 🙇

PR #193, opened 2022-04-10
add some server helper's function test 

TaigaMikami:
I add some server helper's function tests.

### DIFF
--- a/packages/server/src/helpers/convertCertBufferToPEM.test.ts
+++ b/packages/server/src/helpers/convertCertBufferToPEM.test.ts
@@ -1,0 +1,13 @@
+import convertCertBufferToPEM from "./convertCertBufferToPEM";
+
+test('should return pem when input is base64URLString', () => {
+  const input = "Y2VydEJ1ZmZlclN0cmluZw";
+  const actual = convertCertBufferToPEM(input);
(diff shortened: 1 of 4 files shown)


packages/server/src/helpers/convertCertBufferToPEM.test.ts
```diff
@@ -0,0 +1,13 @@
+import convertCertBufferToPEM from "./convertCertBufferToPEM";
+
+test('should return pem when input is base64URLString', () => {
+  const input = "Y2VydEJ1ZmZlclN0cmluZw";
+  const actual = convertCertBufferToPEM(input);
+  expect(actual).toEqual(`-----BEGIN CERTIFICATE-----\nY2VydEJ1ZmZlclN0cmluZw==\n-----END CERTIFICATE-----\n`)
```
MasterKale, line 6: Would you mind adding one more test that ensures the PEM certificate only has 64 characters per line?

MasterKale (changes requested):

MasterKale:
@TaigaMikami Thank you for submitting more tests! These all look good. I just had the one request for one more test and then I think these are ready to merge.

packages/server/src/helpers/convertCertBufferToPEM.test.ts
TaigaMikami: @MasterKale 
Thank you for your review. I fixed it. Could you confirm again?

PR #213, opened 2022-06-15
feat/lerna-upgrade

MasterKale:
This PR bumps Lerna to the latest v5 release, as well as incorporates nx for task running with some basic caching for building and testing.


MasterKale:
I was able to recreate the CI error locally after an `npx lerna clean`, removing node_modules/, running `npm i`, then going into packages/server/ and running `npm run prepublish`.

It turns out server stopped being able to resolve typescript-types. I'll have to figure that out next.

Issue #209, opened 2022-06-10
Support the 'mediation' credential request option 


Manduro:
Right now the browser's `startAuthentication` method only passes the `publicKey` option to `navigator.credentials.get`: https://github.com/MasterKale/SimpleWebAuthn/blob/5ea1b6834b540dad6f733011556989adc66e569e/packages/browser/src/methods/startAuthentication.ts#L40

Besides the `publicKey` option, I'd like to pass the `mediation` option which is documented here: https://w3c.github.io/webappsec-credential-management/#enumdef-credentialmediationrequirement

Especially the `conditional` value will be useful as it enables browsers to present an autofill option in a login form. See https://github.com/w3c/webauthn/wiki/Explainer:-WebAuthn-Conditional-UI. Safari 16 might be the first to support this to autofill passkeys. Some info about this is available here: https://developer.apple.com/videos/play/wwdc2022/10092/?time=1036

MasterKale:
Thanks for opening this, @Manduro. I've known about conditional UI for quite a while now, but until WWDC it still felt like something being incubated in Chrome Canacry. The demonstration of support for it in macOS Ventura established an actual deadline for anything that'd support it. As such I'm now going to prioritize adding support for it.

MasterKale:
I've made some progress on this the last couple of days 👀 

**Chrome Canary 105:**

<img width="364" alt="Screen Shot 2022-06-16 at 10 58 20 PM" src="https://user-images.githubusercontent.com/5166470/174412791-f2684027-365c-46f2-b1fa-b05a3a391617.png">

**Safari in macOS Ventura Beta:**

![Screen Shot 2022-06-17 at 4 18 17 PM](https://user-images.githubusercontent.com/5166470/174412815-03385b1f-248e-4d5e-9b41-ef4fcf56ef36.png)

I think I've got a good idea of how I want to support this feature within **@simplewebauthn/browser**

MasterKale:
Conditional UI is now supported in **@simplewebauthn/browser@5.3.0**. I went with "browser autofill" as a simpler way of communicating what "Conditional UI" gets you.

I've added docs as well to show how to use it:

https://simplewebauthn.dev/docs/packages/browser/#browser-autofill-aka-conditional-ui

PR #214, opened 2022-06-19
feat/conditional-ui

MasterKale:
This PR adds support for upcoming browsers' [conditional UI](https://github.com/w3c/webauthn/wiki/Explainer:-WebAuthn-Conditional-UI) capability to **@simplewebauthn/browser**'s `startAuthentication()`.

To use it, pass in `true` for the new second `useBrowserAutofill` positional argument to `startAuthentication()`, then specify Promise resolution (or rejection) logic to handle the user possibly, eventually choosing a WebAuthn credential via the browser's autofill mechanism:

```html
<head>
  <!-- ... -->
  <script>
    const { startAuthentication } = SimpleWebAuthnBrowser;

    fetch('/generate-authentication-options')
      .then((options) => {
        startAuthentication(options, true)
          .then(authResp => sendToServerForVerificationAndLogin)
          .catch(err => handleError);
      });
  </script>
</head>
```

**Guidance from platform vendors indicates that it is important to initialize WebAuthn's Conditional UI experience as soon as possible.** Placing this logic in `<head>` should give browsers enough time to query authenticators for any discoverable credentials to display to the user. It may also work to delay painting UI For N milliseconds to achieve the same thing in something like a single-page app that would trigger this experience some time after page load. It's still early days for this capability so ymmv.

A new asynchronous `browserSupportsWebAuthnAutofill()` helper is also being added in this PR. This method can be used independently of the call that's made in `startAuthentication()` when initializing browser autofill support.

Existing authentication UI and logic should not need to be refactored. This new "pending" WebAuthn request will be automatically cancelled on a subsequent execution of `startAuthentication()` in, say, an existing click handler that triggers the browser's typical "modal" WebAuthn experience.

Conditional UI is still a nascent capability, but this method should be pretty reliable since the API is largely settled. This new logic has been successfully tested in both **Chrome Canary**[1][2] and **Safari in the macOS Ventura beta**.

## Screenshots

Chrome Canary:

Screen Shot 2022-06-17 at 4 18 17 PM

Safari in the macOS Ventura beta:

FVfczGlUsAI7kYb

Addresses #209.

------

**[1]** The following flag must be specified when launching Chrome Canary to enable Conditional UI:

```
open -a /Applications/Google\ Chrome\ Canary.app --args --enable-features=WebAuthenticationConditionalUI
```

**[2]** I've been informed there's a race condition deep within Chrome Canary that sometimes prevents the autofill from populating with WebAuthn credentials. If they aren't showing up just keep refreshing and eventually they'll appear. This is a temporary issue and will definitely be resolved before Chrome's Conditional UI support lands in Chrome Stable.

### DIFF
--- a/packages/browser/src/helpers/webAuthnAbortService.ts
+++ b/packages/browser/src/helpers/webAuthnAbortService.ts
@@ -1,0 +1,23 @@
+/**
+ * A way to cancel an existing WebAuthn request, for example to cancel a
+ * WebAuthn autofill authentication request for a manual authentication attempt.
+ */
+class WebAuthnAbortService {
+  private controller: AbortController | undefined;
+
+  /**
+   * Prepare an abort signal that will help support multiple auth attempts without needing to
+   * reload the page
+   */
+  createNewAbortSignal() {
+    // Abort any existing calls to navigator.credentials.get()
+    if (this.controller) {
+      this.controller.abort();
+    }
+
+    this.controller = new AbortController();
+    return this.controller.signal;
+  }
+}
+
+export const webauthnAbortService = new WebAuthnAbortService();
--- a/packages/browser/src/helpers/webAuthnAbortService.ts
+++ b/packages/browser/src/helpers/webAuthnAbortService.ts
@@ -18,6 +18,10 @@ class WebAuthnAbortService {
     this.controller = new AbortController();
     return this.controller.signal;
   }
+
+  reset() {
+    this.controller = undefined;
+  }
 }
 
 export const webauthnAbortService = new WebAuthnAbortService();
--- a/packages/browser/src/helpers/webAuthnAbortService.ts
+++ b/packages/browser/src/helpers/webAuthnAbortService.ts
@@ -10,7 +10,7 @@ class WebAuthnAbortService {
    * reload the page
    */
   createNewAbortSignal() {
-    // Abort any existing calls to navigator.credentials.get()
+    // Abort any existing calls to navigator.credentials.create() or navigator.credentials.get()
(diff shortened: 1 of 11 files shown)


packages/browser/src/methods/startAuthentication.ts
```diff
@@ -37,7 +42,39 @@ export default async function startAuthentication(
...
+      throw Error('No <input> with `"webauthn"` in its `autocomplete` attribute was detected');
+    }
+
+    // `CredentialMediationRequirement` doesn't know about "conditional" yet as of
+    // typescript@4.6.3
+    options.mediation = 'conditional' as CredentialMediationRequirement;
+    // Conditional UI requires an empty allow list
+    publicKey.allowCredentials = [];
+
+    console.log('ready for conditional UI');
```
MasterKale, line 71: Remove this before merging

packages/browser/src/helpers/webAuthnAbortService.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+class WebAuthnAbortService {
+  private controller: AbortController | undefined;
+
+  /**
+   * Prepare an abort signal that will help support multiple auth attempts without needing to
+   * reload the page
+   */
+  createNewAbortSignal() {
+    // Abort any existing calls to navigator.credentials.get()
```
MasterKale, line 13: Change this to make it more generic since it'll get used in `startRegistration()` as well

MasterKale:
@P4sca1 good question, let's take this into a Discussion: https://github.com/MasterKale/SimpleWebAuthn/discussions/225

PR #223, opened 2022-07-05
Fix credential ID comparison in example

nflaig:
This PR fixes the `credentialID` comparison used to identify if the device/authenticator the user tries to register already exists.
Since the `credentialID` is of type `Buffer` it is not possible to use `===` as this would always return `false`, instead `Buffer.equals` has to be used which checks if the values have exactly the same bytes.

### DIFF
--- a/example/index.ts
+++ b/example/index.ts
@@ -182,7 +182,7 @@ app.post('/verify-registration', async (req, res) => {
   if (verified && registrationInfo) {
     const { credentialPublicKey, credentialID, counter } = registrationInfo;
 
-    const existingDevice = user.devices.find(device => device.credentialID === credentialID);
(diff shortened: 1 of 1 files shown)


MasterKale:
Haha, nice catch! Can you include a short blurb in your PR description for some context? Otherwise this lgtm :shipit:

nflaig:
@MasterKale I updated the description.

Maybe one more thing related to this, I noticed that authenticators will generate a new `credentialID` anyhow, even if it is for the same server. The only really effective way I found to prevent that a user registers the same authenticator twice is by using `excludeCredentials` when sending the registration options to the client. But I guess it is still good to check the `credentialID` to really ensure that you don't end up with two authenticators with the same `credentialID` which might cause issues/inconsistencies when finding the authenticator provided to `verifyAuthenticationResponse`

PR #230, opened 2022-07-21
Return `AuthenticationExtensionsAuthenticatorOutputs` as part of registration and authentication

agektmr:
This pull request:
* Defines `AuthenticationExtensionsAuthenticatorOutputs`
* Puts extensions in AuthenticatorData as part of output of `verifyRegistrationResponse` and `verifyAuthenticationResponse`.
* Adds tests to parse extensions in AuthenticatorData in `verifyRegistrationResponse` and `verifyAuthenticationResponse`.


packages/server/src/helpers/decodeExtensions.ts
```diff
@@ -0,0 +1,29 @@
+import cbor from 'cbor';
+
+/**
+ * Convert an extension data buffer to a proper object
+ *
+ * @param extensionDataBuffer Extension Data buffer
```
MasterKale, line 6: Can you mention "authenticator data" somewhere in here? Unless I'm reading this wrong this method is specifically about authenticator extensions, not client extensions, right?

packages/server/src/helpers/decodeExtensions.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export default function decodeExtensionDataBuffer(extensionDataBuffer: Buffer): AuthenticationExtensionsAuthenticatorOutputs {
+  const toCBOR: AuthenticationExtensionsAuthenticatorOutputs = cbor.decodeAllSync(extensionDataBuffer)[0];
```
MasterKale, line 9: Have you tried feeding this method a bad buffer? I wonder if a try/catch in here would help with debugging bad data, if only to specify that it was an error while parsing it. Maybe the stack trace would be enough, though 🤔

packages/server/src/helpers/decodeExtensions.ts
```diff
@@ -0,0 +1,29 @@
...
+  return toCBOR;
+}
+
+export type AuthenticationExtensionsAuthenticatorOutputs = {
+  devicePublicKey?: DevicePublicKeyJSON;
+  uvm?: UvmJSON;
```
MasterKale, line 15: Android's going to support `uvm`, eh? 🎉

packages/server/src/helpers/decodeExtensions.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export type DevicePublicKeyJSON = {
+  dpk?: Buffer;
+  scp?: Buffer;
+  sig?: string;
+  aaguid?: Buffer;
+}
+
+// TODO: Need to verify this format
+// https://w3c.github.io/webauthn/#sctn-uvm-extension.
```
MasterKale, line 26: Is this `TODO` still valid?

packages/server/src/registration/verifyRegistrationResponse.test.ts
```diff
@@ -580,6 +580,38 @@ test('should return credential backup info', async () => {
...
+      clientExtensionResults: {},
+      type: 'public-key',
+    },
+    expectedChallenge: 'Arkrl_Fx_MvcJ_eIqCTQ7-tbESISR14-pyPRh0K-1A8',
+    expectedOrigin: 'android:apk-key-hash:gx7sq_pxhxhrIQdLyfG0pxKwiJ7hOk2DJQ4xvKd438Q',
+    expectedRPID: 'try-webauthn.appspot.com',
+  });
+
+  expect(verification.registrationInfo?.extensions).toMatchObject({
+    'devicePublicKey': expect.anything()
```
MasterKale, line 611: Can you know the output and put that in here? This _technically_ doesn't test the actual parsing logic.

packages/server/src/registration/verifyRegistrationResponse.ts
```diff
@@ -132,7 +133,13 @@ export default async function verifyRegistrationResponse(
...
 
   const parsedAuthData = parseAuthenticatorData(authData);
-  const { aaguid, rpIdHash, flags, credentialID, counter, credentialPublicKey } = parsedAuthData;
+  const { aaguid, rpIdHash, flags, credentialID, counter, credentialPublicKey, extensionsDataBuffer } = parsedAuthData;
+
+  let extensions: AuthenticationExtensionsAuthenticatorOutputs = {};
+
+  if (flags.ed && extensionsDataBuffer) {
+    extensions = decodeExtensionDataBuffer(extensionsDataBuffer);
+  }
```
MasterKale, line 142: Can this be moved into `parseAuthenticatorData()` instead? I like what you're doing with returning an empty object if no data is set, but I think it would be cleaner if `parseAuthenticatorData()` parsed the extension data when present too, since there's already other kinds of parsing that goes on in this existing method.

I'm thinking something like this:

```ts
// packages/server/src/helpers/parseAuthenticatorData.ts
type ParsedAuthenticatorData = {
  // ...
  extensionsData: AuthenticationExtensionsAuthenticatorOutputs,
  extensionsDataBuffer: Buffer,
}
```

MasterKale:
@agektmr Thank you for working on this! Many of us are very excited for the DPK extension, it'll be awesome to support it as it becomes available!

packages/server/src/helpers/decodeExtensions.ts
```diff
@@ -0,0 +1,29 @@
+import cbor from 'cbor';
+
+/**
+ * Convert an extension data buffer to a proper object
+ *
+ * @param extensionDataBuffer Extension Data buffer
+ */
+export default function decodeExtensionDataBuffer(extensionDataBuffer: Buffer): AuthenticationExtensionsAuthenticatorOutputs {
```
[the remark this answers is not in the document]
agektmr: Done

packages/server/src/helpers/decodeExtensions.ts
agektmr: Done

packages/server/src/helpers/decodeExtensions.ts
agektmr: Added a dedicated test for exceptional case.

packages/server/src/helpers/decodeExtensions.ts
agektmr: Consider this is just a placeholder.

packages/server/src/helpers/decodeExtensions.ts
agektmr: Yes, it's still a TODO.

packages/server/src/registration/verifyRegistrationResponse.test.ts
agektmr: Done.

packages/server/src/registration/verifyRegistrationResponse.ts
agektmr: Done.

agektmr: Actually, it's a good to return both Buffer and object. Let me fix it.

packages/server/src/helpers/decodeAuthenticatorExtensions.ts
```diff
@@ -0,0 +1,37 @@
...
+export default function decodeAuthenticatorExtensionData(
+  extensionData: Buffer
+): AuthenticationExtensionsAuthenticatorOutputs | undefined {
+  let toCBOR: AuthenticationExtensionsAuthenticatorOutputs | undefined;
+  try {
+    toCBOR = cbor.decodeAllSync(extensionData)[0];
+  } catch (e) {
+    console.error(e);
+    toCBOR = undefined;
+  }
```
MasterKale, line 17: Looking at existing use of `cbor.decodeAllSync(...)[0]`, I think I want this method to go the way of `convertPublicKeyToPEM()`, which throws on bad data but tries to provide some context. For consistency's sake, how about this instead?

```suggestion
  try {
    toCBOR = cbor.decodeAllSync(extensionData)[0];
  } catch (err) {
    const _err = err as Error;
    throw new Error(`Error decoding authenticator extensions: ${_err.message}`);
  }
```

packages/server/src/helpers/decodeAuthenticatorExtensions.ts
```diff
@@ -0,0 +1,37 @@
...
+  return toCBOR;
+}
+
+export type AuthenticationExtensionsAuthenticatorOutputs = {
+  devicePublicKey?: DevicePublicKeyJSON;
+  uvm?: UvmJSON;
+}
+
+export type DevicePublicKeyJSON = {
```
MasterKale, line 26: This should probably be renamed to something like `DevicePublicKeyAuthenticatorOutput` pending a final name decision in https://github.com/w3c/webauthn/pull/1663/. Looking at that PR there's clearly going to be output in authenticator data as well as from `getClientExtensionResults()` (e.g. `DevicePublicKeyClientOutput`) and it'll be useful to be able to distinguish these with unique type names.

agektmr: Agreed!

packages/server/src/helpers/decodeAuthenticatorExtensions.ts
```diff
@@ -0,0 +1,37 @@
...
+export type DevicePublicKeyAuthenticatorOutput = {
+  dpk?: Buffer;
+  scp?: Buffer;
+  sig?: string;
+  aaguid?: Buffer;
+}
+
+// TODO: Need to verify this format
+// https://w3c.github.io/webauthn/#sctn-uvm-extension.
+export type UvmJSON = {
```
MasterKale, line 35: Whoops, one more here - after poring over the spec this morning I'm understanding that CDDL definitions are a little handwave-y on guidance of what these types should be named. Can this get renamed to `UVMAuthenticatorOutput` instead please?

```suggestion
  uvm?: UVMAuthenticatorOutput;
}

export type DevicePublicKeyAuthenticatorOutput = {
  dpk?: Buffer;
  scp?: Buffer;
  sig?: string;
  aaguid?: Buffer;
}

// TODO: Need to verify this format
// https://w3c.github.io/webauthn/#sctn-uvm-extension.
export type UVMAuthenticatorOutput = {
```

agektmr: Done.

packages/server/src/registration/verifyRegistrationResponse.ts
```diff
@@ -288,6 +291,7 @@ export type VerifiedRegistrationResponse = {
     userVerified: boolean;
     credentialDeviceType: CredentialDeviceType;
     credentialBackedUp: boolean;
+    extensionsData?: AuthenticationExtensionsAuthenticatorOutputs;
```
MasterKale, line 294: I hadn't paid all that much attention to extensions until now, so I'm trying to make sure your additions communicate things clearly to the library's users.

In this case I think it'd make sense to rename this to better communicate that these are **authenticator** extension outputs, instead of what comes in as `credential.clientExtensionResults` when `verifyRegistrationResponse()` is called:

```suggestion
    authenticatorExtensionResults?: AuthenticationExtensionsAuthenticatorOutputs;
```

Truth be told I'd rather this be called `authenticatorExtensionOutputs`, but there's already the concept of "Results" from `clientExtensionResults` here so we should probably go with that here too:

https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/typescript-types/src/index.ts#L71

packages/server/src/helpers/decodeAuthenticatorExtensions.ts
```diff
@@ -0,0 +1,37 @@
+import cbor from 'cbor';
+
+/**
+ * Convert authenticator extension data buffer to a proper object
+ *
+ * @param extensionData Authenticator Extension Data buffer
+ */
+export default function decodeAuthenticatorExtensionData(
```
MasterKale, line 8: Ah, I just caught this, I'm slowly migrating this project **away** from `export default` to all-named-exports. Can you rename this to `decodeAuthenticatorExtensions()` and update related imports?

packages/server/src/registration/verifyRegistrationResponse.ts
MasterKale: Oh, and same thing in `verifyAuthenticationResponse()` too

packages/server/src/helpers/decodeAuthenticatorExtensions.ts
agektmr: I've already committed this, but throwing means it needs to be caught by the caller. Do you think that's the way to go? I'd just skip parsing rather than terminating entire verification process. What do you think?

packages/server/src/helpers/decodeAuthenticatorExtensions.ts
agektmr: Done.

packages/server/src/helpers/decodeAuthenticatorExtensions.ts
MasterKale: My current thinking is that if we're told that extension data is present (via `flags.ed`) but it's malformed (improper CBOR, etc...) then I think it makes sense to terminate since it indicates an issue with the authenticator.

agektmr: I think that makes sense!

packages/server/src/registration/verifyRegistrationResponse.ts
```diff
@@ -274,6 +284,7 @@ export default async function verifyRegistrationResponse(
  * @param registrationInfo.credentialBackedUp Whether or not the multi-device credential has been
  * backed up. Always `false` for single-device credentials. **Should be kept in a DB for later
  * reference!**
+ * @param registrationInfo?.extensions The extensions returned by the browser
```
MasterKale, line 287:
```suggestion
 * @param registrationInfo?.authenticatorExtensionResults The extensions returned by the browser
```

packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -218,5 +220,6 @@ export type VerifiedAuthenticationResponse = {
     newCounter: number;
     credentialDeviceType: CredentialDeviceType;
     credentialBackedUp: boolean;
+    authenticatorExtensionResults?: AuthenticationExtensionsAuthenticatorOutputs;
```
MasterKale, line 223: This needs a corresponding entry in the docstring

packages/server/src/registration/verifyRegistrationResponse.ts
```diff
@@ -132,7 +133,15 @@ export default async function verifyRegistrationResponse(
...
   const parsedAuthData = parseAuthenticatorData(authData);
-  const { aaguid, rpIdHash, flags, credentialID, counter, credentialPublicKey } = parsedAuthData;
+  const {
+    aaguid,
+    rpIdHash,
+    flags,
+    credentialID,
+    counter,
+    credentialPublicKey,
+    authenticatorExtensionResults
```
MasterKale, line 143: Argh, the pre-commit hook didn't catch this? I need to investigate why, this should have a trailing comma.

packages/server/src/helpers/parseAuthenticatorData.ts
```diff
@@ -95,5 +100,6 @@ export type ParsedAuthenticatorData = {
   aaguid?: Buffer;
   credentialID?: Buffer;
   credentialPublicKey?: Buffer;
-  extensionsDataBuffer?: Buffer;
+  authenticatorExtensionResults?: AuthenticationExtensionsAuthenticatorOutputs;
+  authenticatorExtensionsDataBuffer?: Buffer;
```
MasterKale, line 104: I may have been a bit too ambiguous with my feedback about `authenticatorExtensionResults`. In my mind these two would remain `extensionsData` and `extensionsDataBuffer` (since it's implied they're authenticator extensions given we're in `parseAuthenticatorData()`), and then `extensionsData` gets returned from `verifyRegistrationResponse()` and `verifyAuthenticationResponse()` as `authenticatorExtensionResults`.

packages/server/src/helpers/parseAuthenticatorData.ts
```diff
@@ -74,7 +78,8 @@ export default function parseAuthenticatorData(authData: Buffer): ParsedAuthenti
     aaguid,
     credentialID,
     credentialPublicKey,
-    extensionsDataBuffer,
+    authenticatorExtensionResults,
+    authenticatorExtensionsDataBuffer
```
MasterKale, line 82: Another nit until I figure out why pre-commit hooks didn't handle this, but can you add a trailing comma here?

MasterKale (commented):
Thank you for enduring all of my feedback, this might just be the last round of it before we can merge this 🤩

MasterKale (approved):

PR #237, opened 2022-08-03
browser package should be a module

### LINKED ISSUE #236: vitejs complains about the browser package
```
@simplewebauthn/browser doesn't appear to be written in CJS, but also doesn't appear to be a valid ES module (i.e. it doesn't have "type": "module" or an .mjs extension for the entry point). Please contact the package author to fix.
```

I don't really know if that is an actual problem.

bomgar:
fixes #236

### DIFF
--- a/packages/browser/jest-environment.js
+++ b/packages/browser/jest-environment.js
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
-const Environment = require('jest-environment-jsdom');
+import Environment from 'jest-environment-jsdom';
 
 /**
  * Set up a custom JSDOM-based test environment for Jest so we can add things JSDOM doesn't support
@@ -13,7 +13,7 @@ class CustomTestEnvironment extends Environment {
      * Solved thanks to https://stackoverflow.com/a/57713960/2133271
      */
     if (typeof this.global.TextEncoder === 'undefined') {
-      const { TextEncoder } = require('util');
+      const { TextEncoder } = await import('util');
       this.global.TextEncoder = TextEncoder;
     }
 
@@ -21,10 +21,10 @@ class CustomTestEnvironment extends Environment {
      * Add support for TextDecoder to JSDOM
      */
     if (typeof this.global.TextDecoder === 'undefined') {
-      const { TextDecoder } = require('util');
+      const { TextEncoder } = await import('util');
       this.global.TextDecoder = TextDecoder;
     }
   }
 }
 
-module.exports = CustomTestEnvironment;
+export default CustomTestEnvironment;
--- a/packages/browser/jest-environment.js
+++ b/packages/browser/jest-environment.js
@@ -21,7 +21,7 @@ class CustomTestEnvironment extends Environment {
      * Add support for TextDecoder to JSDOM
      */
     if (typeof this.global.TextDecoder === 'undefined') {
-      const { TextEncoder } = await import('util');
+      const { TextDecoder } = await import('util');
       this.global.TextDecoder = TextDecoder;
     }
   }
--- a/packages/browser/jest.config.js
+++ b/packages/browser/jest.config.js
@@ -1,7 +1,7 @@
-const rootConfig = require('../../jest.config');
-
-module.exports = {
-  ...rootConfig,
+export default {
+  preset: 'ts-jest',
+  collectCoverageFrom: ['<rootDir>/src/**/*.{js,ts}'],
+  coverageDirectory: 'coverage',
   testEnvironment: '<rootDir>/jest-environment.js',
   setupFilesAfterEnv: ['<rootDir>/src/setupTests.ts'],
 };
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,5 +1,0 @@
-module.exports = {
-  preset: 'ts-jest',
-  collectCoverageFrom: ['<rootDir>/src/**/*.{js,ts}'],
-  coverageDirectory: 'coverage',
-};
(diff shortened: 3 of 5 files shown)


bomgar:
To be honest I don't really know what I'm doing here. I just fixed the errors I got.

packages/browser/jest.config.js
```diff
@@ -1,7 +1,7 @@
-const rootConfig = require('../../jest.config');
```
MasterKale, line 1: I experimented a bit and found out you _can_ use `import` on this:

```ts
import rootConfig from '../../jest.config.mjs';
```

But, as you can see from the path, you need to rename **jest.config.js** in the root of this repo to a `.mjs` file and also convert its `module.exports =` to an `export default`. That'll restore the ability to spread `rootConfig` into the object that's exported from here and avoid the duplication.

packages/browser/jest-environment.js
```diff
@@ -13,18 +14,16 @@ class CustomTestEnvironment extends Environment {
      * Solved thanks to https://stackoverflow.com/a/57713960/2133271
      */
     if (typeof this.global.TextEncoder === 'undefined') {
-      const { TextEncoder } = require('util');
```
MasterKale, line 16: You can use `await import()` here instead of moving this import to the top:

```ts
const { TextEncoder } = await import('util');
```

packages/browser/jest-environment.js
```diff
@@ -13,18 +14,16 @@ class CustomTestEnvironment extends Environment {
...
       this.global.TextEncoder = TextEncoder;
     }
 
     /**
      * Add support for TextDecoder to JSDOM
      */
     if (typeof this.global.TextDecoder === 'undefined') {
-      const { TextDecoder } = require('util');
```
MasterKale, line 24: Just like with `TextEncoder` you can use `await import()` here to preserve the importing of `TextDecoder`:

```ts
const { TextDecoder } = await import('util');
```

MasterKale (changes requested):
@bomgar Thank you for the submission. Just a few small tweaks and I'd say this is ready to go. Thank you for including the changes to the test configs.

bomgar:
@MasterKale i did what change the file to `mjs` but now the server import doesn't work anymore. This is annoying stuff to avoid copy pasting 3 lines.

bomgar:
@MasterKale I changed the server to type `module`, too. Now the build is green. Is this ok?

packages/browser/jest-environment.js
```diff
@@ -13,18 +13,18 @@ class CustomTestEnvironment extends Environment {
...
+      const { TextEncoder } = await import('util');
       this.global.TextEncoder = TextEncoder;
     }
 
     /**
      * Add support for TextDecoder to JSDOM
      */
     if (typeof this.global.TextDecoder === 'undefined') {
-      const { TextDecoder } = require('util');
+      const { TextEncoder } = await import('util');
```
MasterKale, line 24: This needs to be Text**Decoder**

```suggestion
      const { TextDecoder } = await import('util');
```

MasterKale:
> @MasterKale I changed the server to type `module`, too. Now the build is green. Is this ok?

I did a before and after build of **server** after adding `"type": "module"` to its package.json, and the diff says nothing changed so I thought it'd be fine to add that:

Screen Shot 2022-08-03 at 5 13 22 PM

However it appears adding that breaks projects that use `require()`, as I observed in the example project when I tried using it:

```
/Users/matt/Developer/simplewebauthn/example/index.ts:43
const app = express();
                 ^
example/node_modules/@cspotcode/source-map-support/source-map-support.js:590
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/matt/Developer/simplewebauthn/packages/server/dist/index.js from /Users/matt/Developer/simplewebauthn/example/index.ts not supported.
index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /Users/matt/Developer/simplewebauthn/packages/server/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

I wonder now if it's worth trying to preserve the deduplication of the Jest config...

Tell you what, can you actually get rid of **jest.config.mjs**, and just copy its three config lines into the **browser** and **server** jest configs? Then forget about adding `"type": "module"` to **server**'s package.json, that's not something this project can afford to break right now.

P4sca1:
This PR updates the server package to `"type": "module"`, which is wrong, as it is currently compiled into a CJS module.

It is a good idea, to convert the package to pure ESM. Therefore the `tscondig.json` file needs to be updated, so that valid ESM is produced. This would be a breaking change, since the module could no longer be `require()`'d.
This gist provides very helpful information around this topic: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

An alternative approach would be to produce both an ESM bundle and a CJS bundle and use [conditional exports](https://nodejs.org/api/packages.html#conditional-exports). [esbuild](https://esbuild.github.io) could be used instead of rollup to do the job. I have a similar build pipeline here for reference: https://github.com/P4sca1/cron-schedule/tree/v3.0.6.

Personally, I would go with the ESM only approach, as it is easier to setup and maintain. This is also what other libraries did in the past - it is time to move away from CJS to ESM. This should then be a major release, which should maybe include the gist I referenced earlier.

MasterKale:
**Server** currently works fine in both ESM and CJS codebases, so I don't immediately see the need to make any changes to it. That's why I mention forgetting about adding `"type": "module"` to its package.json.

We can explore the possibility of refactoring server's build but it's definitely outside the scope of this PR.

bomgar:
@MasterKale I made the adjustments.

MasterKale (approved):

MasterKale:
Thank you for your contribution, @bomgar!

Issue #238, opened 2022-08-05
A handful of server errors coming from Windows devices


BrunoBernardino:
Hi @MasterKale ! We're using SimpleWebAuth at https://github.com/padloc/padloc ([server](https://github.com/padloc/padloc/blob/main/packages/server/src/auth/webauthn.ts) and [client](https://github.com/padloc/padloc/blob/main/packages/app/src/lib/auth/webauthn.ts)) and we've been seeing quite a few errors for Windows devices, failing authentication. These vary, but always fall into one of these two:

1.
```
Error: PubArea unique is not same as public key x and y (TPM|ECC)

    at verifyAttestationTPM (/padloc/packages/server/node_modules/@simplewebauthn/server/src/registration/verifications/tpm/verifyAttestationTPM.ts:115:13)

    at verifyRegistrationResponse (/padloc/packages/server/node_modules/@simplewebauthn/server/src/registration/verifyRegistrationResponse.ts:223:42)

    at WebAuthnServer.activateAuthenticator (/padloc/packages/server/src/auth/webauthn.ts:116:80)
```

2.
```
Stack Trace:

Error: Public key algorithm [object Object] did not match any metadata algorithms [] (TPM)

    at verifyAttestationTPM (/padloc/packages/server/node_modules/@simplewebauthn/server/src/registration/verifications/tpm/verifyAttestationTPM.ts:267:13)

    at runMicrotasks ()

    at processTicksAndRejections (node:internal/process/task_queues:96:5)

    at verifyRegistrationResponse (/padloc/packages/server/node_modules/@simplewebauthn/server/src/registration/verifyRegistrationResponse.ts:223:16)

    at WebAuthnServer.activateAuthenticator (/padloc/packages/server/src/auth/webauthn.ts:116:48)
```

We're wondering if there's something we should do differently, or if these are a symptom of something else (like Windows Hello)?

Thanks!

FYI @MaKleSoft

MasterKale:
@BrunoBernardino Interesting! Can you share a couple of responses with those problematic TPM attestation statements? And would have any insights into the OS that they're being generated on?

MasterKale:
> Error: PubArea unique is not same as public key x and y (TPM|ECC)

[...]

PR #239, opened 2022-08-05
fix/tpm-pub-area-parsing

MasterKale:
This PR updates how a TPM attestation statement's public area gets parsed. I hadn't actually seen ECC responses till now, but with one in hand I was able to enhance this library's support for them.

This should fix the following errors reported in #238:

```
Error: PubArea unique is not same as public key x and y (TPM|ECC)
```

@simplewebauthn/server was also incorrectly comparing the TPM public area's ECC curve ID to the COSE public key `crv` parameter; that issue is fixed here as well.


aseigler (approved):
LGTM!

PR #262, opened 2022-08-19
Implement a component that verifies Device Public Keys

agektmr:
This pull request implements a component that verifies DPKs following [the current specification](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html#sctn-device-publickey-extension).

On [registration](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html#sctn-device-publickey-extension-verification-create):
- The RP obtains a DPK output.
- Verify it with `verifyDpkSignature`.
- Store the DPK to the database.

On [authentication](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html#sctn-device-publickey-extension-verification-get):
- The RP obtains a DPK output.
- Verify it with `verifyDpkSignature`.
- Retrieve stored DPKs with the same credential ID from the database.
- Compare the DPK with the stored DPKs using `verifyDevicePublicKey`
    - Store the DPK to the database if some conditions meet

This is still WIP hence just a draft at this point.


packages/server/src/extensions/devicePubKey.ts
```diff
@@ -0,0 +1,156 @@
...
+    return false;
+  }
+  if (!devicePubKey.scope.equals(expectedDPK.scope)) {
+    return false;
+  }
+  if (devicePubKey.fmt !== expectedDPK.fmt) {
+    return false;
+  }
+  if (devicePubKey.attStmt ||
+      expectedDPK.attStmt) {
```
MasterKale, line 119: Would it make sense for this block to be `if (devicePubKey.attStmt && expectedDPK.attStmt) {...}`? The `return false;` on L123 when either is missing if either is present is hard to grok.

In fact maybe there should be two more `if ()` statements handling those two combinations explicitly:

- `if (devicePubKey.attStmt && !expectedDPK.attStmt) {...}`
- `if (!devicePubKey.attStmt && expectedDPK.attStmt) {...}`

And what about this case? It's unclear whether it should return true or false:

- `if (!devicePubKey.attStmt && !expectedDPK.attStmt) {...}`

MasterKale (commented):
Thank you for the submission @agektmr! Leaving some preliminary notes, I'm going to try and look more closely at things over the weekend so I might have a couple more comments coming your way.

packages/server/src/extensions/devicePubKey.ts
agektmr: True. This looks unnecessarily complex. Fixing.

packages/server/src/extensions/devicePubKey.ts
```diff
@@ -0,0 +1,156 @@
...
+ * 3. Verify that `signature` is a valid signature over the assertion signature
+ *    input by the device public key *dpk*. (The signature algorithm is the same
+ *    as for the user credential.)
+ * @param clientDataJSON 
+ * @param authData 
+ * @param devicePubKey
+ * @param signature 
+ * @returns Promise<boolean>
+ */
+export async function verifyDpkSignature(
```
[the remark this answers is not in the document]
agektmr: Fixed.

packages/server/src/extensions/devicePubKey.ts
```diff
@@ -0,0 +1,156 @@
...
+  clientDataJSON: string,
+  authData: Buffer,
+  devicePubKey: DevicePublicKeyAuthenticatorOutput,
+  signature: Buffer,
+): Promise<boolean> {
```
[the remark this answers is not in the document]
agektmr: Fixed.

packages/server/src/extensions/devicePubKey.ts
```diff
@@ -0,0 +1,156 @@
...
+ *    *attObjForDevicePublicKey* fields, then perform binary equality checks
+ *    between the corresponding stored values and the extracted field values.
+ * @param devicePubKey 
+ * @param expectedDPK 
+ * @returns boolean
+ */
+export function verifyDevicePublicKey(
+  devicePubKey: DevicePublicKeyAuthenticatorOutput,
+  expectedDPK: DevicePublicKeyAuthenticatorOutput
+): boolean {
```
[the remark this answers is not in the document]
agektmr: 👍

packages/server/src/extensions/devicePubKey.ts
agektmr: Fixed.

agektmr:
Reorganized the interface a bit.

BTW, at this stage I only implemented the logic as standalone functions -- verify the signature and verify the DPK (binary equality check) in order to be invoked by the RP assuming it requires database access. However, I realized `verifyDpkSignature` don't necessarily be called by the RP and can only be called by `verifyRegistrationResponse` and `verifyAuthenticationResponse`. Let's read the spec closely and make the decision as we move forward.

packages/server/src/extensions/devicePublicKey/verifyDevicePublicKey.ts
```diff
@@ -0,0 +1,69 @@
...
+ *    *attObjForDevicePublicKey* fields, then perform binary equality checks
+ *    between the corresponding stored values and the extracted field values.
+ * @param devicePubKey 
+ * @param expectedDPK 
+ * @returns boolean
+ */
+export function verifyDevicePublicKey(
+  devicePubKey: DevicePublicKeyAuthenticatorOutput,
+  expectedDPK: DevicePublicKeyAuthenticatorOutput
+): boolean {
```
MasterKale, line 17: In your PR description you imagine the following RP actions during authentication:

> - Retrieve stored DPKs with the same credential ID from the database.
> - Compare the DPK with the stored DPKs using `verifyDevicePublicKey`

Are you imagining an RP that cares about DPK performing something like this (please forgive the pseudocode)?

```js
const verification = await verifyAuthenticationResponse({...});

const userDPKs = await DPKService.getDPKsForUser(user.id);
let foundDPKMatch = false;
for (const userDPK of userDPKs) {
  foundDPKMatch = verifyDevicePublicKey(verification.dpk, userDPK);

  if (foundDPKMatch) {
    break;
  }
}

if (!foundDPKMatch) {
  throw new Error('User tried authenticating from an unknown device');
}
```

Based on your response I might propose an alternative solution, I wonder if something like this couldn't get added into `verifyAuthenticationResponse()` instead 🤔

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    response: {
+      clientDataJSON: 'eyJ0eXBlIjoid2ViYXV0aG4uY3JlYXRlIiwiY2hhbGxlbmdlIjoiNFF3dmdUVjhlaTF5TzYyaDF2bVFNWFM2SDZ3UGc0YWt2eDNKcF9VakF3cyIsIm9yaWdpbiI6ImFuZHJvaWQ6YXBrLWtleS1oYXNoOmd4N3NxX3B4aHhocklRZEx5ZkcwcHhLd2lKN2hPazJESlE0eHZLZDQzOFEiLCJhbmRyb2lkUGFja2FnZU5hbWUiOiJjb20uZmlkby5leGFtcGxlLmZpZG8yYXBpZXhhbXBsZSJ9',
+      attestationObject: 'o2NmbXRkbm9uZWdhdHRTdG10oGhhdXRoRGF0YVkBaQ11_MVj_ad52y40PupImIh1i3hUnUk6T9vqHNlqoxzExQAAAAAAAAAAAAAAAAAAAAAAAAAAABBzGMMHWHmcbqOlBbcR55k0pQECAyYgASFYIIukb9t-EtGUOa2t6YiJEAgz7GyqBN4DFTCzkcMiUGqIIlggmm6GzBPSzP9IYJnX-89R_zmKl6-qQSeQ2qomEC6Cr32hbGRldmljZVB1YktleaVjZHBrWE2lAQIDJiABIVgg7erT_TV2nCPTQN3Bgwp_8g5zVfKdHHWqDcK2rBgup9MiWCA0UdyZkq-UaCW0QZRfydE04Xtzql_qlYA1HnyT9dNlE2NzaWdYRjBEAiALHwldj84eCsg9f0fHD9hylpUK8N_TGOBKBQPoNvfWjQIgChIUfdbO1HBavxbZGQxIt4v23FqFLB2nzwip-avJ6etlbm9uY2VAZXNjb3BlQQBmYWFndWlkUAAAAAAAAAAAAAAAAAAAAAA=',
+    },
+    id: 'cxjDB1h5nG6jpQW3EeeZNA',
+    rawId: 'cxjDB1h5nG6jpQW3EeeZNA',
+    type: 'public-key',
+    transports: [],
+    clientExtensionResults: ''
+  } as RegistrationCredentialJSON;
```
MasterKale, line 15: `RegistrationCredentialJSON` should be assigned as the type of `credential` instead of using `as` afterwards:

```suggestion
  const credential: RegistrationCredentialJSON = {
    response: {
      clientDataJSON: 'eyJ0eXBlIjoid2ViYXV0aG4uY3JlYXRlIiwiY2hhbGxlbmdlIjoiNFF3dmdUVjhlaTF5TzYyaDF2bVFNWFM2SDZ3UGc0YWt2eDNKcF9VakF3cyIsIm9yaWdpbiI6ImFuZHJvaWQ6YXBrLWtleS1oYXNoOmd4N3NxX3B4aHhocklRZEx5ZkcwcHhLd2lKN2hPazJESlE0eHZLZDQzOFEiLCJhbmRyb2lkUGFja2FnZU5hbWUiOiJjb20uZmlkby5leGFtcGxlLmZpZG8yYXBpZXhhbXBsZSJ9',
      attestationObject: 'o2NmbXRkbm9uZWdhdHRTdG10oGhhdXRoRGF0YVkBaQ11_MVj_ad52y40PupImIh1i3hUnUk6T9vqHNlqoxzExQAAAAAAAAAAAAAAAAAAAAAAAAAAABBzGMMHWHmcbqOlBbcR55k0pQECAyYgASFYIIukb9t-EtGUOa2t6YiJEAgz7GyqBN4DFTCzkcMiUGqIIlggmm6GzBPSzP9IYJnX-89R_zmKl6-qQSeQ2qomEC6Cr32hbGRldmljZVB1YktleaVjZHBrWE2lAQIDJiABIVgg7erT_TV2nCPTQN3Bgwp_8g5zVfKdHHWqDcK2rBgup9MiWCA0UdyZkq-UaCW0QZRfydE04Xtzql_qlYA1HnyT9dNlE2NzaWdYRjBEAiALHwldj84eCsg9f0fHD9hylpUK8N_TGOBKBQPoNvfWjQIgChIUfdbO1HBavxbZGQxIt4v23FqFLB2nzwip-avJ6etlbm9uY2VAZXNjb3BlQQBmYWFndWlkUAAAAAAAAAAAAAAAAAAAAAA=',
    },
    id: 'cxjDB1h5nG6jpQW3EeeZNA',
    rawId: 'cxjDB1h5nG6jpQW3EeeZNA',
    type: 'public-key',
    transports: [],
    clientExtensionResults: ''
  };
```

MasterKale: In general please opt for the `const foo: FooType` pattern instead of casting with `as`.

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.ts
```diff
@@ -0,0 +1,132 @@
...
+ *    input by the device public key *dpk*. (The signature algorithm is the same
+ *    as for the user credential.)
+ * @param options 
+ * @returns Promise<boolean>
+ */
+export async function verifyDpkSignature(
+  options: VerifyDpkSignatureOpts
+): Promise<boolean> {
+  const { credential, devicePubKey, signature } = options;
+
```
MasterKale, line 40: Below I see a mix of using unpacking to access properties on `credential`, and directly accessing properties like `credential.response`. Consider refactoring these to a single unpacking here of all the values you want off of `credential`:

```js
const { id, response } = credential;
```

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.ts
```diff
@@ -0,0 +1,132 @@
...
+  let authData;
+  if (isAuthenticationResponse(credential)) {
+    const { authenticatorData } = credential.response;
+    authData = base64url.toBuffer(authenticatorData);
+  }
+  if (isRegistrationResponse(credential)) {
+    const attestationObject = base64url.toBuffer(credential.response.attestationObject);
+    const { authData: authenticatorData } = decodeAttestationObject(attestationObject);
+    authData = authenticatorData;
+  }
```
MasterKale, line 50: Authentications happen more often than registrations, so we can probably completely skip the `isRegistrationResponse()` the majority of the time with an `else if` here. Allowing both type guards to execute when we know one will always be false can probably be considered "inefficient":

```suggestion
  if (isAuthenticationResponse(credential)) {
    const { authenticatorData } = credential.response;
    authData = base64url.toBuffer(authenticatorData);
  } else if (isRegistrationResponse(credential)) {
    const attestationObject = base64url.toBuffer(credential.response.attestationObject);
    const { authData: authenticatorData } = decodeAttestationObject(attestationObject);
    authData = authenticatorData;
  }
```

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.ts
```diff
@@ -0,0 +1,132 @@
...
+  if (authData === undefined) {
+    throw new Error("AuthenticatorResponse doesn't include authenticatorData");
+  }
+
+  const { clientDataJSON } = credential.response;
+  const clientDataHash = toHash(base64url.toBuffer(clientDataJSON));
+  const { id } = credential;
+  const credentialID = base64url.toBuffer(id);
```
MasterKale, line 58: These could be squashed down to a single line to make things a little neater here:
```suggestion
  const credentialID = base64url.toBuffer(credential.id);
```

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.ts
```diff
@@ -0,0 +1,132 @@
...
+  const { rpIdHash } = parseAuthenticatorData(authData);
+  
+  if (!credentialID) {
+    // `authData` without `credentialID` can't be verified.
+    return false;
+  }
```
MasterKale, line 64: How might `credentialID` ever be falsey here? 🤔

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  const nonce = devicePubKey.nonce ? devicePubKey.nonce : Buffer.from('');
+  // According to the spec, `authData` and `clientDataHash` are concatenated as
+  // the signature base, but this is an interim implementation.
+  const signatureBase = Buffer.concat([credentialID, clientDataHash, nonce]);
+  const credentialPublicKey = devicePubKey.dpk;
```
MasterKale, line 70: Can you please leave a note here that we're assigning a "device public key" to the variable name "`credentialPublicKey`" because "something something some methods below expect to get passed the credential's public key, but in this context the DPK needs to be used instead"?

It wasn't immediately obvious why it got assigned a seemingly conflicting name, but then seeing where the value gets passed in later it all made sense. We'll probably thank ourselves later if we leave a quick note like that now.

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.ts
```diff
@@ -0,0 +1,132 @@
...
+      verified = true;
+    } else {
+      throw new Error(`Unsupported Attestation Format: ${devicePubKey.fmt}`);
+    }
+    if (!verified) {
+      return false;
+    }
+  } else {
+    // `fmt` and `attStmt` are required fields, though at the moment, as the test
+    // attestation omits it. Ignore this case intentionally.
```
MasterKale, line 115: > ...though at the moment, as the test attestation omits it.

Does this mean this needs a TODO to revisit this case in the future? It's ambiguous if the "test attestation" intends to always include `fmt` and `attStmt`, or if we'll always need to handle them potentially being undefined. Are they optional as per the current draft of the spec?

packages/server/src/extensions/devicePublicKey/verifyDevicePublicKey.ts
MasterKale: What if we add a `userDevicePubKeys?: DevicePublicKeyAuthenticatorOutput[]` argument to `verifyAuthenticationResponse()`, and then include the logic within `verifyAuthenticationResponse()` to test each of the user's existing devicePubKeys against the authentication response's verified DPK? Most RP's might not need to wire up any additional DPK checks that way, which I _think_ reduces the need to export `verifyDevicePublicKey()` at all.

Leveraging DPK is kinda complex; I want to continue to make WebAuthn as **simple** to use as possible, so we should consider alternative approaches like this.

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.ts
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html#sctn-device-publickey-extension-verification-create
+ * 3. Verify that `signature` is a valid signature over the assertion signature
+ *    input by the device public key *dpk*. (The signature algorithm is the same
+ *    as for the user credential.)
+ * @param options 
+ * @returns Promise<boolean>
+ */
+export async function verifyDpkSignature(
```
MasterKale, line 36: I neglected to mention this earlier: can this get renamed to `verifyDevicePublicKeySignature()` please?

I'd like to see "DPK" spelled out in methods like this that might eventually become part of the public API, so that users just coming into this functionality have one fewer acronym to juggle.

MasterKale: Ah, the `VerifyDpkSignatureOpts` type should get spelled out too

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.ts
```diff
@@ -0,0 +1,132 @@
...
+  devicePubKey: DevicePublicKeyAuthenticatorOutput;
+  signature: Buffer;
+};
+
+/**
+ * https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html#sctn-device-publickey-extension-verification-create
+ * 3. Verify that `signature` is a valid signature over the assertion signature
+ *    input by the device public key *dpk*. (The signature algorithm is the same
+ *    as for the user credential.)
+ * @param options
```
MasterKale, line 33: TODO: Flesh out these docs. One thing I'd like to see clarified is where `signature` will come from. I believe the latest discussions on where `signature` will reside is within `clientExtensionResults` as part of a new "unsigned extension output", because the signature :

> **Client extension output**
> 
> An ArrayBuffer containing the [=device public key attestation object=] that was returned in the authenticator extension output and the signature returned as the [=unsigned extension output=].

I pulled this from the current draft of the DPK PR: https://github.com/w3c/webauthn/pull/1663/files#diff-5e793325cd2bfc452e268a4aa2f02b4024dd9584bd1db3c2595f61f1ecf7b985R6805

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.test.ts
agektmr: Of course, yours is better :)

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.ts
agektmr: This is a left over from older code. I'll ditch it.

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.ts
agektmr: That's a good idea. I'll add.

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.ts
agektmr: Makes sense. I will fix them.

packages/server/src/extensions/devicePublicKey/verifyDpkSignature.ts
agektmr: In the sample payload I use, `devicePubKey` includes the signature as `sig` (I think it came from older spec). Also, currently the `devicePubKey` can be only passed through authenticator extension output and not client extension output (I wasn't aware of "unsigned extension output" and I will need to read the discussion).
So I'm expecting some updates to the sample payload but I'd like to proceed with what we have for the moment.

packages/server/src/extensions/devicePublicKey/verifyDevicePublicKey.ts
agektmr: +1! I made these independent because I wasn't sure if I could aim to release this code as part of SimpleWebAuthn as a beta branch. With your help, I'm happy to make these a part of `verifyAuthenticationResonse`.

Will need to read the spec better before I can make the change though.

packages/server/src/extensions/devicePublicKey/verifyDevicePublicKey.ts
```diff
@@ -0,0 +1,69 @@
...
+ * 4. If the Relying Party's user account mapped to the *credential*.id in play
+ *    (i.e., for the user being authenticated) holds `aaguid`, `dpk`, `scope`,
+ *    `fmt`, and `attStmt` values corresponding to the extracted
+ *    *attObjForDevicePublicKey* fields, then perform binary equality checks
+ *    between the corresponding stored values and the extracted field values.
+ * @param devicePubKey 
+ * @param expectedDPK 
+ * @returns boolean
+ */
+export function verifyDevicePublicKey(
```
MasterKale, line 14: I just thought up a potentially better name for this method: `isRecognizedDevice()`, with arguments `responseDevicePublicKey` and `knownDevicePublicKey`? Something like that.

`verifyPublicKey()` kept bugging me because this isn't the same kind of verification method as `verifyDevicePublicKeySignature()`; I think the name of this method should reflect the fact that this is really how RP's will be able to tell when the DPK represents a new device vs one that's been registered before.

agektmr:
I think this is ready for your review now.

MasterKale:
While I'm reviewing, can you try rebasing this on the latest **beta** branch and force pushing up? For some reason this PR thinks it needs to include changes to things like **CHANGELOG.md** and some package.json files. But these changes were made in commits on **beta** like **b0aa153a** and **d467da8f** that are already up on origin from when I last synced beta with the latest changes from master.

packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -251,8 +289,13 @@ export type VerifiedAuthenticationResponse = {
   authenticationInfo: {
     credentialID: Buffer;
     newCounter: number;
+    userVerified: boolean;
     credentialDeviceType: CredentialDeviceType;
     credentialBackedUp: boolean;
-    authenticatorExtensionResults?: AuthenticationExtensionsAuthenticatorOutputs;
   };
+  extensionOutputs: ExtensionOutputs;
```
MasterKale, line 296: I'd like to see this included within `authenticationInfo` like the other values, as `authenticationInfo` has been a bucket of information that I intend RPs to only be interested in if `verified === true`.

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
```diff
@@ -0,0 +1,140 @@
...
+      if (!responseDevicePublicKey.scope.equals(userDPK.scope)) {
+        return false;
+      }
+      return true;
+    });
+
+    // more than one match
+    if (matchedDPKs.length > 1) {
+      throw new Error('It is undetermined whether this is a known device.');
+    }
```
MasterKale, line 42: I'm curious why this is a requirement. If a duplicate is found, why not simply skip and move on?

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
```diff
@@ -0,0 +1,140 @@
...
+      const index = userDevicePublicKeys.findIndex(userDPK => {
+        return responseDevicePublicKey.dpk.equals(userDPK.dpk)
+      });
+      if (index === -1) {
+        // If `attObjForDevicePublicKey.dpk` did not match any of the Relying
+        // Party's stored dpk values for this user account and `credential.id`
+        // pair then:
+        const isValidDPKAttestation = await verifyDevicePublicKeyAttestation(responseDevicePublicKey);
+        if (!isValidDPKAttestation) {
+          throw new Error('Device public key attestation is invalid.');
```
MasterKale, line 57: Alternative wording:

> DevicePublicKey attestation could not be verified

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    // zero matches
+    if (matchedDPKs.length === 0) {
+      // This is possibly a new device.
+
+      const index = userDevicePublicKeys.findIndex(userDPK => {
+        return responseDevicePublicKey.dpk.equals(userDPK.dpk)
+      });
+      if (index === -1) {
```
MasterKale, line 51: Unless I'm mistaken, `index` will always be `-1` here because `matchedDPKs` needs to be empty to get here. And the reason `matchedDPKs` would be empty here is because the call to `.filter()` above already filtered out items in `userDevicePublicKeys` with `dpk` properties that didn't match `responseDevicePublicKey.dpk`. Thus `.findIndex()` will never find any matches in `userDevicePublicKeys` here, right? 🤔

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
```diff
@@ -0,0 +1,140 @@
...
+    // key with a valid signature and thus a known device. Terminate these
+    // verification steps.
+    if (responseDevicePublicKey.fmt && responseDevicePublicKey.fmt !== 'none') {
+      // Perform a binary equality check of `attStmt`.
+      const knownAttStmt = matchedDPKs[0].attStmt;
+      if (!checkAttStmtBinaryEquality(responseDevicePublicKey.attStmt, knownAttStmt)) {
+        // Otherwise, verify attestation
+        const isValidDPKAttestation = await verifyDevicePublicKeyAttestation(responseDevicePublicKey);
+        if (!isValidDPKAttestation) {
+          throw new Error('Device Public Key attestation is invalid.');
```
MasterKale, line 83: Alternative wording:

> DevicePublicKey attestation could not be verified

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  } else {
+    // Otherwise, the Relying Party does not have `attObjForDevicePublicKey`
+    // fields presently mapped to this user account and credential.id pair:
+    const isValidDPKAttestation = await verifyDevicePublicKeyAttestation(responseDevicePublicKey);
+    if (!isValidDPKAttestation) {
+      throw new Error('Device Public Key attestation is invalid.');
+    }
+    return responseDevicePublicKey;
+  }
```
MasterKale, line 98: To clarify, this is the logic that executes if `fmt` is `"none"`? If so can you make a mention of that somewhere in the comments here?

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+export function checkAttStmtBinaryEquality(
+  responseAttStmt?: AttestationStatement,
+  knownAttStmt?: AttestationStatement
```
MasterKale, line 103: We're going with "recognized" in methods like `isRecognizedDevice()`. Maybe `knownAttStmt` can become `recognizedDPKAttStmt` instead?

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
```diff
@@ -0,0 +1,140 @@
...
+    // fields presently mapped to this user account and credential.id pair:
+    const isValidDPKAttestation = await verifyDevicePublicKeyAttestation(responseDevicePublicKey);
+    if (!isValidDPKAttestation) {
+      throw new Error('Device Public Key attestation is invalid.');
+    }
+    return responseDevicePublicKey;
+  }
+}
+
+export function checkAttStmtBinaryEquality(
```
MasterKale, line 101: There are a _lot_ of circumstances in which `attStmt` might return lead to `false` being returned from this method.

For sake of troubleshooting issues with DPK as it evolves, what if this method threw errors explaining which check failed (e.g. "Response DPK sig and Recognized SPK sig did not match", to keep messages concise for all the checks that either is defined before the equivalency check)? Then calls to it can become `try/catch`'s that can append the error message from `checkAttStmtBinaryEquality()`:

```ts
let dpksAreBinaryEqual = false;
try {
  dpksAreBinaryEqual = checkAttStmtBinaryEquality(responseDevicePublicKey.attStmt, knownAttStmt)
} catch (err) {
  const _err = err as Error;
  throw new Error(`DevicePublicKey attStmt's were not equal: ${_err.message}`);
}

if (!dpksAreBinaryEqual) {
  // ...
}

return responseDevicePublicKey;
```

We could then generate more useful errors like this:

> DevicePublicKey attStmt's were not equal: Response DPK sig and Recognized SPK sig did not match

packages/server/src/registration/verifyRegistrationResponse.ts
```diff
@@ -192,6 +194,41 @@ export async function verifyRegistrationResponse(
...
+      }
+
+      // Optionally verify device public key attestation here as per
+      // 10.2.2.3.1. step 4 in the spec.
+      const attResult = await verifyDevicePublicKeyAttestation(devicePubKey);
+      if (!attResult) {
+        throw new Error('Invalid device public key attestation.');
+      }
+    }
+  }
```
MasterKale, line 230: Non-blocking: This is the same logic as in `verifyAuthenticationResponse()`, perhaps we can revisit this logic later and figure out how to pull it out into something reusable?

packages/server/src/index.ts
```diff
@@ -16,6 +18,7 @@ export {
   verifyAuthenticationResponse,
   MetadataService,
   SettingsService,
+  isRecognizedDevice,
```
MasterKale, line 21: Non-blocking: This is fine for a beta release, but my plan with extension helpers like this is to expose them as exports from `@simplewebauthn/server/extensions` (like I do with `@simplewebauthn/server/helpers`)

packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -193,6 +197,39 @@ export async function verifyAuthenticationResponse(
...
+    // was requested in extension inputs. See 7.1 step 18 in the spec.
+
+    // DevicePublicKey sample currently provides the data through authenticator
+    // extension results. 
+    if (extensionsData?.devicePubKey) {
+      const { devicePubKey } = extensionsData;
+      const { sig: signature } = devicePubKey;
+
+      if (!signature) {
+        throw new Error('DevicePublicKey signature is missing.');
```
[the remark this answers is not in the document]
agektmr: `sig` is an old specification and `signature` is supposed to be part of client extension as per client extension output part of https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html#sctn-device-publickey-extension-definition. How about this?:

> DevicePublicKey was missing signature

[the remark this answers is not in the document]
MasterKale: That sounds good to me!

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
agektmr: If you look at [the spec](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html#sctn-device-publickey-extension-verification-get): "4" > "exactly one match"

> Otherwise, check attObjForDevicePublicKey’s attStmt by performing a binary equality check between the corresponding stored and extracted attStmt values. If the result is:

This is where the above check occurs. There's is a continuation:

"unsuccessful"

> Optionally, if attestation was requested and the RP wishes to verify it, verify that attStmt is a correct [attestation statement](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html#attestation-statement), conveying a valid [attestation signature](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html#attestation-signature), by using the [attestation statement format](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html#attestation-statement-format) fmt’s [verification procedure](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html#verification-procedure) given attStmt. See [§ 10.2.2.2.1 Attestation calculations](https://pr-preview.s3.amazonaws.com/w3c/webauthn/pull/1663.html#sctn-device-publickey-attestation-calculations).

Throwing upon `checkAttStmtBinaryEquality()` seems too early to me.

```
      let dpksAreBinaryEqual = false;
      try {
        dpksAreBinaryEqual = checkAttStmtBinaryEquality(responseDevicePublicKey.attStmt, recognizedDPKAttStmt);
      } catch (err) {
        const _err = err as Error;
        // How do we message the error cause?
        // throw new Error(`DevicePublicKey attStmt's were not equal: ${_err.message}`);
        // Otherwise, verify attestation
        const isValidDPKAttestation = await verifyDevicePublicKeyAttestation(responseDevicePublicKey);
        if (!isValidDPKAttestation) {
          throw new Error('DevicePublicKey attestation could not be verified.');
        }
      }
```

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
agektmr: `matchedDPKs` is checking if all `aaguid`, `dpk` and `scope` match. It's possible that *only* `dpk` matches but others are different. My understanding is this section is to determine such case.

packages/server/src/index.ts
agektmr: This function is no longer needed to be exposed.

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
MasterKale: Oh, I see, `matchedDPKs` captures this part of DPK verification:

>  If the Relying Party 's user account mapped to the credential . id in play (i.e., for the user being authenticated) holds aaguid , dpk and scope values corresponding to the extracted attObjForDevicePublicKey fields, then perform binary equality checks between the corresponding stored values and the extracted field values. The Relying Party MAY have more than one set of { aaguid , dpk , scope } values mapped to the user account and credential . id pair and each set MUST be checked.

And this `.findIndex()` attempts to address this part of DPK verification:

>  - zero matches
>  This is possibly a new device public key signifying a new device:
>    - If attObjForDevicePublicKey . dpk did not match any of the Relying Party 's stored dpk values for this user account and credential . id pair then...

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
MasterKale: Alright, you've convinced me. Let's roll with this for now.

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
MasterKale: This is at least accurate to the spec, so I retract my proposal lol

packages/server/src/index.ts
```diff
@@ -8,6 +8,7 @@ import { generateAuthenticationOptions } from './authentication/generateAuthenti
 import { verifyAuthenticationResponse } from './authentication/verifyAuthenticationResponse';
 import { MetadataService } from './services/metadataService';
 import { SettingsService } from './services/settingsService';
+import { DevicePublicKeyAuthenticatorOutput } from './helpers/decodeAuthenticatorExtensions';
```
MasterKale, line 11: This should be an `import type` now since we're no longer importing functionality as well. It should get moved down with the other `import type`'s below too.

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
MasterKale: The spec is pretty clear about this:

> - If the above set of binary equality checks resulted in:
> **more than one match**
>    - Some form of error has occurred. It is indeterminate whether this is a known device. Terminate these verification steps.

packages/server/src/index.ts
MasterKale: Actually this is something that can be done later, I'll leave this unresolved for now so I have something to remind me later.

MasterKale:
Okay, just about everything has been resolved to my satisfaction. Unless you want me to hold off I think we can merge this and cut a beta release.

MasterKale (approved):

MasterKale:
This functionality is available in **@simplewebauthn/server@6.3.0-alpha.0**

PR #267, opened 2022-09-10
Add a better Device Public Key support

agektmr:
This pull request adds a better device public key support.
You may start reviewing this, but I will keep this as draft until the spec is closer to final.


MasterKale (changes requested):
For some reason this diff includes **package-lock.json** files despite their being an entry in **.gitignore** to leave them out. Libraries intentionally omit these files so that consuming libraries are free to upgrade to the latest version of libraries as per the individual libraries' **package.json** files.

Can you please remove all package-lock.json files in **packages/** from this diff?

packages/typescript-types/src/index.ts
```diff
@@ -87,7 +87,34 @@ export interface AuthenticationCredentialJSON
...
+  devicePubKey?: AuthenticationExtensionsDevicePublicKeyOutputs
+}
+
+export interface AuthenticationExtensionsDevicePublicKeyOutputsJSON {
+  authenticatorOutput: Base64URLString;
+  signature: Base64URLString;
+}
+
+export interface AuthenticationExtensionsClientOutputsJSON extends AuthenticationExtensionsClientOutputs {
+  devicePubKey?: AuthenticationExtensionsDevicePublicKeyOutputsJSON
```
MasterKale, line 117: **@simplewebauthn/browser**'s `startRegistration()` and `startAuthentication()` are going to need corresponding updates to encode `devicePubKey` if it's present. Right now I think the library naively calls `credential.getClientExtensionResults()` which, until `devicePubKey`, returned JSON-serializable values that required no additional processing.

packages/typescript-types/src/index.ts
```diff
@@ -87,7 +87,34 @@ export interface AuthenticationCredentialJSON
   extends Omit<AuthenticationCredential, 'response' | 'rawId' | 'getClientExtensionResults'> {
   rawId: Base64URLString;
   response: AuthenticatorAssertionResponseJSON;
-  clientExtensionResults: AuthenticationExtensionsClientOutputs;
+  clientExtensionResults: AuthenticationExtensionsClientOutputsJSON;
+}
+
+export interface AuthenticationExtensionsDevicePublicKeyInputs {
+  attestation: string;
```
MasterKale, line 94: This should be updated to type `AttestationConveyancePreference`, importable with the other entries from `"./dom"`.

packages/typescript-types/src/index.ts
```diff
@@ -87,7 +87,34 @@ export interface AuthenticationCredentialJSON
...
+  attestationFormats: string[];
```
MasterKale, line 95: This can be refined to `AttestationFormat[]`, but unfortunately `AttestationFormat` containing these various values currently resides in **packages/server/src/helpers/decodeAttestationObject.ts**. I suggest we move `AttestationFormat` out of there and into **@simplewebauthn/typescript-types** so that we can share it for use here too.

MasterKale, line 95: The way I'm reading the spec, these should be marked optional because they're not marked `required` in the spec draft:

https://pr-preview.s3.amazonaws.com/w3c/webauthn/1663/f903606...d671894.html#dictdef-authenticationextensionsdevicepublickeyinputs

packages/typescript-types/src/index.ts
```diff
@@ -87,7 +87,34 @@ export interface AuthenticationCredentialJSON
...
+
+export interface AuthenticationExtensionsClientOutputsFuture extends AuthenticationExtensionsClientOutputs {
+  devicePubKey?: AuthenticationExtensionsDevicePublicKeyOutputs
+}
+
+export interface AuthenticationExtensionsDevicePublicKeyOutputsJSON {
+  authenticatorOutput: Base64URLString;
+  signature: Base64URLString;
+}
+
```
MasterKale, line 115: Can this get bumped up above `AuthenticationExtensionsClientOutputsFuture`, to be underneath its corresponding `AuthenticationExtensionsDevicePublicKeyOutputs`? It might make it easier to remember to update one when the other gets updated.

MasterKale (commented):
Can you also revert the changes made to **package-lock.json**? I think whatever created those other package-lock.json files you removed also edited the root one, and looking over the changes it looks like some kind of hoisting took place that shouldn't have happened.

Do you have an idea of what might have done this? Can you think of any commands you ran that might be responsible?

agektmr:
Hmm, no idea. I only run npm scripts.

packages/typescript-types/src/index.ts
agektmr: Done.

packages/browser/src/helpers/parseClientExtensionResults.ts
```diff
@@ -0,0 +1,32 @@
...
+    } else if (key === 'devicePubKey') {
+      const { devicePubKey } = clientExtensionResults;
+      if (!devicePubKey) continue;
+      const authenticatorOutput = bufferToBase64URLString(devicePubKey.authenticatorOutput);
+      const signature = bufferToBase64URLString(devicePubKey.signature);
+      clientExtensionResultsJSON.devicePubKey = { authenticatorOutput, signature };
+    } else if (key === 'uvm') {
+      clientExtensionResultsJSON.uvm = clientExtensionResults.uvm;
+    }
+  }
```
MasterKale, line 29: What about something like this, it's fewer iterations and is a little easier to read IMO:

```suggestion
  const { appid, credProps, devicePubKey, uvm } = clientExtensionResults;

  if (appid) {
    clientExtensionResultsJSON.appid = appid;
  }

  if (credProps) {
    clientExtensionResultsJSON.credProps = credProps;
  }

  if (uvm) {
    clientExtensionResultsJSON.uvm = clientExtensionResults.uvm;
  }

  if (devicePubKey) {
    clientExtensionResultsJSON.devicePubKey = {
      authenticatorOutput: bufferToBase64URLString(devicePubKey.authenticatorOutput),
      signature: bufferToBase64URLString(devicePubKey.signature),
    };
  }
```

packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -197,36 +204,35 @@ export async function verifyAuthenticationResponse(
...
 
-  const extensionOutputs: ExtensionOutputs = {};
+  const extensionOutputs: AuthenticationExtensionOutputs = {};
 
   if (flags.ed) {
-    if (!extensionsData && !clientExtensionResults) {
+    if (!clientExtensionResults) {
       throw new Error('Authenticator data indicated extension data was present,'+
-        ' but no client or authenticator extension data were found');
+        ' but no client extension data were found');
```
MasterKale, line 212: I actually think this line should drop "client" for "authenticator" since my reading of https://www.w3.org/TR/webauthn-2/#authenticator-data tells me that only authenticator extensions are present in `authenticatorData`.
```suggestion
        ' but no authenticator extension data were found');
```

packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -197,36 +204,35 @@ export async function verifyAuthenticationResponse(
...
 
     // DevicePublicKey sample currently provides the data through authenticator
-    // extension results.
-    if (extensionsData?.devicePubKey) {
-      const { devicePubKey } = extensionsData;
-      const { sig: dpkSig } = devicePubKey;
+    // extension results. 
+    if (clientExtensionResults.devicePubKey) {
+      const devicePubKey =  decodeDevicePubKey(clientExtensionResults.devicePubKey);
+      const { authenticatorOutput: encodedAuthenticatorOutput, signature } = devicePubKey;
```
MasterKale, line 222: Why does `authenticatorOutput` get assigned to a different name here? I don't see any other variables by that name in this scope 🤔

packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -197,36 +204,35 @@ export async function verifyAuthenticationResponse(
     }
   }
 
-  const extensionOutputs: ExtensionOutputs = {};
+  const extensionOutputs: AuthenticationExtensionOutputs = {};
 
   if (flags.ed) {
-    if (!extensionsData && !clientExtensionResults) {
+    if (!clientExtensionResults) {
```
MasterKale, line 210: I actually think `!clientExtensionResults` here should be `!extensionsData`, since the `ed` flag in authenticatorData is about the presence of _authenticator_ extension output.

packages/server/src/extensions/devicePublicKey/isRecognizedDevice.ts
```diff
@@ -90,9 +90,6 @@ export async function isRecognizedDevice(
           throw new Error('DevicePublicKey attestation could not be verified.');
```
MasterKale, line 90: This `Error` should reveal why the equality check failed:
```suggestion
          const _err = err as Error;
          throw new Error(`DevicePublicKey attestation could not be verified: ${_err.message}`);
```

packages/server/src/extensions/devicePublicKey/verifyDevicePublicKeyAttestation.ts
```diff
@@ -22,14 +12,14 @@ export async function verifyDevicePublicKeyAttestation(
   devicePubKey: DevicePublicKeyAuthenticatorOutput
 ): Promise<boolean> {
   const { fmt } = devicePubKey;
-  if (fmt === undefined || fmt === 'none') {
+  if (fmt === 'none') {
     return true;
-  }
+  } else {
+    throw new Error('Attestation verification on a device public key is not implemented yet.');
```
MasterKale, line 18: Curious why this might be? Is the attestation statement format not yet decided? I thought it was going to be similar to the attestation statement returned from direct attestation during credential creation.

packages/server/src/extensions/devicePublicKey/verifyDevicePublicKeySignature.ts
```diff
@@ -53,63 +44,11 @@ export async function verifyDevicePublicKeySignature(
...
-  const credentialID = base64url.toBuffer(credential.id);
-  const { rpIdHash } = parseAuthenticatorData(authData);
   
-  const nonce = devicePubKey.nonce ? devicePubKey.nonce : Buffer.from('');
-  // According to the spec, `authData` and `clientDataHash` are concatenated as
-  // the signature base, but this is an interim implementation.
-  const signatureBase = Buffer.concat([credentialID, clientDataHash, nonce]);
+  const signatureBase = Buffer.concat([authData, clientDataHash]);
   // It's a device public key and not a credential public key, but to align with
   // the `verifySinature` signature, we name it `credentialPublicKey`.
```
MasterKale, line 50: These comments can be deleted since the `credentialPublicKey` they're describing is getting deleted here.

packages/server/src/registration/verifyRegistrationResponse.ts
```diff
@@ -194,38 +196,44 @@ export async function verifyRegistrationResponse(
...
   }
 
+  const extensionOutputs: RegistrationExtensionOutputs = {};
+
   if (flags.ed) {
-    if (!extensionsData && !clientExtensionResults) {
-      throw new Error('Extension results are not included despite the flag.');
+    if (!clientExtensionResults) {
+      throw new Error('Authenticator data indicated extension data was present,'+
+        ' but no client extension data were found');
```
MasterKale, line 204: Same thing as in `verifyAuthenticationResponse()`, I'm understanding it to be better to check that `extensionsData` is present since `getClientExtensionResults()` can return different extension results than what the authenticator might include in authenticatorData. Right? Or maybe I've misunderstood `authenticatorData` extensions this whole time.

packages/server/src/authentication/verifyAuthenticationResponse.ts
agektmr: Good catch. Will fix.

packages/server/src/authentication/verifyAuthenticationResponse.ts
agektmr: I wasn't aware that `ed` indicates the presence of authenticator extension output. In that case, we don't need to check both `ed` and `extentionsData` as they are not dealt anyway.

packages/server/src/extensions/devicePublicKey/verifyDevicePublicKeyAttestation.ts
agektmr: I'm not familiar with the reasoning for this, but we are not shipping attestation support for dpk at least the initial version.

packages/server/src/extensions/devicePublicKey/verifyDevicePublicKeySignature.ts
agektmr: What do you mean by "getting deleted"?

packages/server/src/authentication/verifyAuthenticationResponse.ts
MasterKale: Yeah, I'm looking at a Touch ID response in Chrome and `ed` is false even though I requested the client extension `credProps` and got a response back:

https://debugger.simplewebauthn.dev/?credential=ewogICJpZCI6ICJURW0yaXRRUlh6NEEyZ012bDczSHdQMC1pVWZFZzN6VWpVMzBGdk12UklBIiwKICAicmF3SWQiOiAiVEVtMml0UVJYejRBMmdNdmw3M0h3UDAtaVVmRWczelVqVTMwRnZNdlJJQSIsCiAgInJlc3BvbnNlIjogewogICAgImF0dGVzdGF0aW9uT2JqZWN0IjogIm8yTm1iWFJrYm05dVpXZGhkSFJUZEcxMG9HaGhkWFJvUkdGMFlWaWsyeW00U2lwMWNQVkd1UGp0cDR3Z1dvc0VOTU00R3dJMGhTTzBJc1EyNU9SRkFBQUFBSzNPQUFJMXZNWUtaSXNMSmZId1ZRTUFJRXhKdG9yVUVWOC1BTm9ETDVlOXg4RDlQb2xIeElOODFJMU45QmJ6TDBTQXBRRUNBeVlnQVNGWUlOcjJjeFhPLXZ5b0E4ZjVkcVVuMXkzRXV4dVFvamw0MjZPUjUwUjFCdjdJSWxnZzJaOGVPQTRibFY0ZVJrRUlsMllTcEI5TXl5VkJ4M090M2xmUUJ2aXRNQnciLAogICAgImNsaWVudERhdGFKU09OIjogImV5SjBlWEJsSWpvaWQyVmlZWFYwYUc0dVkzSmxZWFJsSWl3aVkyaGhiR3hsYm1kbElqb2lMVlY2TlhsMmJtTm9ORWt6VlROMGVXbzJSVTVtVFZkc1dsUkJhWE16T1ZKc2NtNUVNRkYwZERsV05DSXNJbTl5YVdkcGJpSTZJbWgwZEhCek9pOHZaWGhoYlhCc1pTNXphVzF3YkdWM1pXSmhkWFJvYmk1a1pYWWlMQ0pqY205emMwOXlhV2RwYmlJNlptRnNjMlY5IgogIH0sCiAgInR5cGUiOiAicHVibGljLWtleSIsCiAgImNsaWVudEV4dGVuc2lvblJlc3VsdHMiOiB7CiAgICAiY3JlZFByb3BzIjogewogICAgICAicmsiOiB0cnVlCiAgICB9CiAgfSwKICAidHJhbnNwb3J0cyI6IFsKICAgICJpbnRlcm5hbCIKICBdCn0

So based on this I'd say `ed` only reflects whether `authenticatorData` contains any extension data.

packages/server/src/extensions/devicePublicKey/verifyDevicePublicKeySignature.ts
MasterKale: Oh, whoops, I saw all the red and missed that you removed all that to assign `authenticatorOutput.dpk` to `credentialPublicKey`.

packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -197,36 +204,30 @@ export async function verifyAuthenticationResponse(
     }
   }
 
-  const extensionOutputs: ExtensionOutputs = {};
-
-  if (flags.ed) {
-    if (!extensionsData && !clientExtensionResults) {
-      throw new Error('Authenticator data indicated extension data was present,'+
-        ' but no client or authenticator extension data were found');
-    }
```
MasterKale, line 206: Note to self: make sure this doesn't end up breaking FIDO Conformance

packages/server/src/registration/verifyRegistrationResponse.ts
```diff
@@ -194,38 +196,39 @@ export async function verifyRegistrationResponse(
     throw new Error(`Unexpected public key alg "${alg}", expected one of "${supported}"`);
   }
 
-  if (flags.ed) {
-    if (!extensionsData && !clientExtensionResults) {
-      throw new Error('Extension results are not included despite the flag.');
-    }
```
MasterKale, line 200: Note to self: make sure the removal of this check doesn't end up breaking FIDO Conformance before merging into master

MasterKale (approved):

MasterKale:
This changes have been published as **@simplewebauthn/server@6.3.0-alpha.1** and **@simplewebauthn/browser@6.3.0-alpha.1**

PR #299, opened 2022-11-11
feat/isomorphic

MasterKale:
This PR refactors SimpleWebAuthn to work in more runtimes than just Node. This is generally achieved by refactoring away references to Node-specific data classes, like `Buffer`, and referencing Web APIs in a way that should support use of **@simplewebauthn/server** in more non-Node runtimes than before.

Non-isomorphic third-party dependencies have also been replaced, with efforts taken to prevent breaking changes to external API inputs and outputs.

## Noteworthy changes:

- The project now targets **Node 16** and above
- Existing projects that pass in `Buffer` for method arguments that are now marked `Uint8Array` should not break, due to [Node's `Buffer` being a subclass of `Uint8Array`](https://nodejs.org/api/buffer.html)

### Registration

- The following values returned from `verifyRegistrationResponse()` are now a `Uint8Array` instead of a `Buffer`
  - `aaguid`
  - `authData`
  - `clientDataHash`
  - `credentialID`
  - `credentialPublicKey`
  - `rpIdHash`

### Authentication 

- The following values returned from `verifyAuthenticationResponse()` are now a `Uint8Array` instead of a `Buffer`
  - `credentialID`

### Helpers (`@simplewebauthn/server/helpers`)

- There are several new helpers for working with WebAuthn-related data types that should work in all runtimes:
  - `isoCBOR` for working with CBOR-encoded values
  - `isoCrypto` for leveraging the WebCrypto API when working with various WebAuthn/FIDO2 data structures
  - `isoBase64URL` for encoding and decoding values into base64url (with optional base64 support)
  - `isoUint8Array` for working with `Uint8Array`s
  - `cose` for working with COSE-related methods and types
- The `isBase64URLString()` helper is now `isoBase64URL.isBase64url()`
- The `decodeCborFirst()` helper is now `isoCBOR.decodeFirst()`
- The `convertPublicKeyToPEM()` helper has been removed


packages/server/package.json
```diff
@@ -30,10 +30,11 @@
     "access": "public"
   },
   "engines": {
-    "node": ">=14.0.0"
+    "node": ">=16.0.0"
   },
   "scripts": {
     "build": "rimraf dist && tsc",
+    "build:lerna-debug": "rimraf dist && tsc > output.txt; cat output.txt; rm output.txt",
```
MasterKale, line 37: This helped me troubleshoot `tsc` errors which Lerna was swallowing because it didn't output `stdout` on a non-0 exit code. See https://github.com/lerna/lerna/issues/3384

packages/server/src/helpers/iso/isoCrypto/digest.ts
```diff
@@ -0,0 +1,25 @@
+import { webcrypto } from 'node:crypto';
+
+import { COSEALG } from '../../cose';
+import { mapCoseAlgToWebCryptoAlg } from './mapCoseAlgToWebCryptoAlg';
+
+/**
+ * Generate a digest of the provided data.
+ *
+ * @param data The data to generate a digest of
+ * @param algorithm A COSE algorithm ID that maps to the SHA algorithm. Default: `-7` (for SHA-256)
```
MasterKale, line 10: Update this to remove the mention of a default value

packages/server/src/helpers/iso/isoCrypto/normalizeSHAAlgorithm.ts
```diff
@@ -0,0 +1,13 @@
+import { SubtleCryptoAlg } from './structs';
+
+/**
+ * Convert algorithms like "SHA1", "sha256", etc... into values like "SHA-1", "SHA-256", etc...
+ * that `.digest()` will accept
+ */
+export function normalizeSHAAlgorithm(algorithm: string): SubtleCryptoAlg {
```
MasterKale, line 7: Double check that this method is used anymore

packages/server/src/helpers/iso/isoCrypto/structs.ts
```diff
@@ -0,0 +1,3 @@
+export type SubtleCryptoAlg = "SHA-1" | "SHA-256" | "SHA-384" | "SHA-512";
```
MasterKale, line 1: Figure out why the linter won't lint these files

packages/server/src/helpers/iso/isoCrypto/verifyEC2.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  if (shouldRemoveLeadingZero(rBytes)) {
+    rBytes = rBytes.slice(1);
+  }
+
+  if (shouldRemoveLeadingZero(sBytes)) {
+    sBytes = sBytes.slice(1);
+  }
+
+  const finalSignature = isoUint8Array.concat([rBytes, sBytes]);
```
MasterKale, line 35: This should get moved under the public key `.get()` calls. No point in doing this if the public key is malformed.

packages/server/src/helpers/iso/isoCrypto/verifyRSA.ts
```diff
@@ -0,0 +1,104 @@
+import WebCrypto from '@simplewebauthn/iso-webcrypto';
+
+import { COSEALG, COSEKEYS, COSEPublicKeyRSA, isCOSEAlg } from '../../cose';
+import { mapCoseAlgToWebCryptoAlg } from './mapCoseAlgToWebCryptoAlg';
+import { importKey } from './importKey';
+import { isoBase64URL } from '../index';
+import { mapCoseAlgToWebCryptoKeyAlgName } from './mapCoseAlgToWebCryptoKeyAlgName';
+
+/**
+ *
```
MasterKale, line 10: TODO: Actually write this

PR #320, opened 2022-12-28
feat/webauthn-L3-json-types

MasterKale:
This PR updates the **typescript-types**, **browser**, and **server** packages with new JSON serialization-related types and functions as defined in L3 of the WebAuthn specification.

The new [`PublicKeyCredential.parseCreationOptionsFromJSON()`](https://w3c.github.io/webauthn/#sctn-parseCreationOptionsFromJSON), [`PublicKeyCredential.parseRequestOptionsFromJSON()`](https://w3c.github.io/webauthn/#sctn-parseRequestOptionsFromJSON), and [`PublicKeyCredential.toJSON()`](https://w3c.github.io/webauthn/#dom-publickeycredential-tojson) methods have been added to **typescript-types**, but  **browser** package will _not_ make use of them.

The "simpler" use of SimpleWebAuthn will always be to pair **browser** with **server**, if an RP doesn't want to think about it. However, because of the way in which SimpleWebAuthn methods are being updated in this PR, RP's that choose to eschew use of **browser** for these JSON methods should gradually become able to spin their own front end logic with the official API methods and still use the values and methods from **server** for options and verification.

## Breaking Changes

- The `RegistrationCredentialJSON` type has been replaced by the `RegistrationResponseJSON` type
- The `AuthenticationCredentialJSON` type has been replaced by the `AuthenticationResponseJSON` type
- `RegistrationResponseJSON.transports` has been relocated into `RegistrationCredentialJSON.response.transports` to mirror response structure in the WebAuthn spec
- **@simplewebauthn/server**'s `verifyRegistrationResponse()` method has had its `credential` argument renamed to `response`
- **@simplewebauthn/server**'s `verifyAuthenticationResponse()` method has had its `credential` argument renamed to `response`


packages/server/src/registration/verifyRegistrationResponse.ts
```diff
@@ -30,7 +30,7 @@ import { verifyAttestationAndroidKey } from './verifications/verifyAttestationAn
 import { verifyAttestationApple } from './verifications/verifyAttestationApple';
 
 export type VerifyRegistrationResponseOpts = {
-  credential: RegistrationCredentialJSON;
+  credential: RegistrationResponseJSON;
```
MasterKale, line 33: I should probably rename `credential` to `response`...

packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -15,7 +15,7 @@ import { matchExpectedRPID } from '../helpers/matchExpectedRPID';
 import { isoUint8Array, isoBase64URL } from '../helpers/iso';
 
 export type VerifyAuthenticationResponseOpts = {
-  credential: AuthenticationCredentialJSON;
+  credential: AuthenticationResponseJSON;
```
MasterKale, line 18: I should probably rename credential to response...

Issue #325, opened 2023-01-05
Question: Use autofill before identifying users


Radiergummi:
I'm working on implementing SimpleWebAuthn for a new app I'm working on, and there's one thing I can't wrap my head around. The client side autofill UI requires authenticator options to determine which of its local keys the server will accept.  
The server, as implemented, requires the user ID however:
```typescript
app.get('/generate-authentication-options', (req, res) => {
  // You need to know the user by this point
  const user = inMemoryUserDeviceDB[loggedInUserId];
```
_(See [example/index.ts#L207-L209](https://github.com/MasterKale/SimpleWebAuthn/blob/3dec91f6fbdfba201e253436f07786a6ef7f99bf/example/index.ts#L207-L209))_

At this point, I'm trying to offer users to use their pass key instead of continuing with their email address - so I really _can't_ have them signed in already. How am I supposed to generate authentication options in that situation?

I suppose the example app is just a little convoluted, and remembers the user ID in global server state, so it will stick around after the registration is complete; unless apps have a single user, however, that approach won't work for real-life applications.  
I _think_ I'd need to pass the client options for autofill if no user ID was included in the request, and serve the full assertion challenge on subsequent requests including a user ID, but I'm not quite sure how to do that securely, or if I understood something fundamentally wrong.  
Do you have some advice?

MasterKale:
As a matter of fact someone asked that same question earlier this week, take a look at this discussion and my response in it and let me know if it answers your question:

https://github.com/MasterKale/SimpleWebAuthn/discussions/321#discussioncomment-4551790

If it does, then I'll make this issue about _finally_ adding this guidance to the docs, so that more people might be able to discover them when implementing passkeys.

MasterKale:
I added a new section to the Passkeys docs with suggestions on how to manage challenges during registration and authentication. Give it a look and let me know what you think:

https://simplewebauthn.dev/docs/advanced/passkeys/#remembering-challenges

Radiergummi:
That's a splendid paragraph and explains things very well. This will surely help others in the future. Thank you very much for all your work, @MasterKale!

PR #326, opened 2023-01-05
Add sessions to the example app

Radiergummi:
This PR adds sessions to the example app, which can be used to store the current challenge in. This provides better guidance to users trying to implement passkey flows using SimpleWebAuthn. Ideally, the passkey docs could be expanded a little to bring the general concept across - once you know how it works, it's immediately clear the ephemeral challenge will need to be stored separately, and sessions are a great fit for that. 

This will solve #103, #325, and aid users landing in discussion #321.

I hope this helps you a little, @MasterKale :)

### DIFF
--- a/example/index.ts
+++ b/example/index.ts
@@ -54,15 +54,20 @@ const {
 
 app.use(express.static('./public/'));
 app.use(express.json());
-app.use( session( {
-  secret: 'secret123',
-  saveUninitialized: true,
-  resave: false,
-  cookie: { maxAge: 86400000 },
-  store: new MemoryStore( {
-    checkPeriod: 86_400_000 // prune expired entries every 24h
-  } ),
-} ) );
+app.use(
+  session({
+    secret: 'secret123',
+    saveUninitialized: true,
+    resave: false,
+    cookie: {
+      maxAge: 86400000,
+      httpOnly: true, // Ensure to not expose session cookies to clientside scripts
+    },
+    store: new MemoryStore({
+      checkPeriod: 86_400_000, // prune expired entries every 24h
+    }),
+  }),
+);
 
 /**
  * If the words "metadata statements" mean anything to you, you'll want to enable this route. It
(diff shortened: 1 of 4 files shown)


example/index.ts
```diff
@@ -51,6 +54,15 @@ const {
 
 app.use(express.static('./public/'));
 app.use(express.json());
+app.use( session( {
+  secret: 'secret123',
+  saveUninitialized: true,
+  resave: false,
+  cookie: { maxAge: 86400000 },
```
MasterKale, line 61: I took a quick look at `express-session` and see that it defaults to setting HTTP-only cookies. Can you explicitly set `httpOnly: true` here anyway to reinforce the use of more secure cookies for this kind of session tracking? It's something I'd suggest to others and want to make sure that gets captured here.

MasterKale:
@Radiergummi Thank you for the PR! I like this upgrade to the example project, not too complex but really helps clarify how RP's can leverage sessions for tracking expected challenges.

Aside from my one request I think this is good to go.

example/index.ts
```diff
@@ -51,6 +54,15 @@ const {
...
+  store: new MemoryStore( {
+    checkPeriod: 86_400_000 // prune expired entries every 24h
+  } ),
+} ) );
```
MasterKale, line 65: I still haven't figured out why my precommit hooks aren't auto-formatting things, but would you mind running `npm run lint` from the root of the repo and confirm it formats this code? If not you should be able to run this manually from the root of the repo instead:

```
prettier --write example/**/*.ts && eslint --fix example/**/*.ts
```

example/index.ts
Radiergummi: Sure, I added the option 👍

example/index.ts
Radiergummi: I couldn't get eslint to work properly, sadly - will get to that tomorrow.

MasterKale: That's fine, looking at the formatting here after you latest commit I think it looks like the formatter might have run somehow anyway. It's _good enough_ now 😄

MasterKale:
lgtm :shipit:

PR #367, opened 2023-03-01
feat/better-errors

### LINKED ISSUE #357: Better error type identification in the browser
## Describe the issue

When catching errors thrown by the browser API, it's desirable to identify what type of error occurred so that the user experience can be altered accordingly.

For example, did the user cancel the authentication? Was the browser unable to fulfill the request for some reason? Based on the error thrown, the user could be prompted to try again, or in some cases (e.g., browser autofill init), just ignore the error altogether.

This would also help cases where autofill _and_ explicit `startAuthentication()` requests are used in parallel: one to help a user leverage an existing passkey even if they're not familiar with the terminology and another for users who know what they want immediately. The autofill request might be intentionally cancelled due to a more explicit request being initiated.

Currently, the library throws generic `Error`s, but it would be useful to customize those errors and set some sort of `code` property that can be programmatically relied upon to not change, even though english string messaging might. (e.g., `code: E_CANCELLED` or `code: E_NO_INPUTS`)

I am happy to PR changes in this regard if its agreed that this would be a helpful addition.

### SimpleWebAuthn Libraries

```sh
$ npm list --depth=0 | grep @simplewebauthn
├── @simplewebauthn/browser@7.0.1
├── @simplewebauthn/typescript-types@7.0.0
```

MasterKale:
This PR augments identified errors raised during calls to `startRegistration()` and `startAuthentication()` with the following:

- Original errors are available as the [MDN-official `cause` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause)
- Distinct, constant error codes are available as a new `code` property with the following possible values:
  - `'ERROR_CEREMONY_ABORTED'`
  - `'ERROR_INVALID_DOMAIN'`
  - `'ERROR_INVALID_RP_ID'`
  - `'ERROR_INVALID_USER_ID_LENGTH'`
  - `'ERROR_MALFORMED_PUBKEYCREDPARAMS'`
  - `'ERROR_AUTHENTICATOR_GENERAL_ERROR'`
  - `'ERROR_AUTHENTICATOR_MISSING_DISCOVERABLE_CREDENTIAL_SUPPORT'`
  - `'ERROR_AUTHENTICATOR_MISSING_USER_VERIFICATION_SUPPORT'`
  - `'ERROR_AUTHENTICATOR_PREVIOUSLY_REGISTERED'`
  - `'ERROR_AUTHENTICATOR_NO_SUPPORTED_PUBKEYCREDPARAMS_ALG'`
  - `'ERROR_PASSTHROUGH_SEE_CAUSE_PROPERTY'`

Fixes #357


packages/browser/src/helpers/identifyAuthenticationError.ts
```diff
@@ -20,32 +20,47 @@ export function identifyAuthenticationError({
   if (error.name === 'AbortError') {
     if (options.signal === new AbortController().signal) {
       // https://www.w3.org/TR/webauthn-2/#sctn-createCredential (Step 16)
-      return new WebAuthnError('Authentication ceremony was sent an abort signal', 'AbortError');
+      return new WebAuthnError({
+        message: 'Authentication ceremony was sent an abort signal',
+        code: 'ERROR_CEREMONY_ABORTED',
```
mhamann, line 25: Since you're using TypeScript, any reason not to leverage an exported `enum` here so that consumers can do things like `if (err.code === ErrorCode.ERROR_CEREMONY_ABORTED)`?

MasterKale: I lean towards union types since it allows for this:

```ts
if (err.code === 'ERROR_CEREMONY_ABORTED') {
  // ...
}
```

In my experience TypeScript has been good about suggesting autocompletion of the strings in the union type without requiring any additional imports.

mhamann: Gotcha. 👍🏻

PR #415, opened 2023-07-19
Add origin and rpID in authentication and registration info

### LINKED ISSUE #414: Add rpID and origin used in the authenticationInfo and registrationInfo
## Describe the proposal

During registration and authentication, metadata info are returned by the "verify" function.
I'd like to suggest to add the rpID and origin being used as the library accept an array of origin and rpID. 
This will be useful in order to let developers add custom logic or logging.

See here: 
- https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/server/src/registration/verifyRegistrationResponse.ts#L238
- https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/server/src/authentication/verifyAuthenticationResponse.ts#L211

Mikescops:
For now I only added the origin to the response.
The rpID is hashed, I'm not sure it is super relevant, what do you think @MasterKale ?

Closes #414

### DIFF
--- a/packages/server/src/authentication/verifyAuthenticationResponse.test.ts
+++ b/packages/server/src/authentication/verifyAuthenticationResponse.test.ts
@@ -44,6 +44,7 @@ test('should return authenticator info after verification', async () => {
 
   expect(verification.authenticationInfo.newCounter).toEqual(144);
   expect(verification.authenticationInfo.credentialID).toEqual(authenticator.credentialID);
+  expect(verification.authenticationInfo?.origin).toEqual(assertionOrigin);
 });
 
 test('should throw when response challenge is not expected value', async () => {
@@ -224,6 +225,7 @@ test('should support multiple possible origins', async () => {
   });
 
   expect(verification.verified).toEqual(true);
+  expect(verification.authenticationInfo?.origin).toEqual(assertionOrigin);
 });
 
 test('should throw an error if origin not in list of expected origins', async () => {
@@ -372,6 +374,7 @@ test('should return credential backup info', async () => {
 
   expect(verification.authenticationInfo?.credentialDeviceType).toEqual('singleDevice');
   expect(verification.authenticationInfo?.credentialBackedUp).toEqual(false);
+  expect(verification.authenticationInfo?.origin).toEqual(assertionOrigin);
 });
 
 /**
--- a/packages/server/src/authentication/verifyAuthenticationResponse.test.ts
+++ b/packages/server/src/authentication/verifyAuthenticationResponse.test.ts
@@ -45,6 +45,7 @@ test('should return authenticator info after verification', async () => {
   expect(verification.authenticationInfo.newCounter).toEqual(144);
   expect(verification.authenticationInfo.credentialID).toEqual(authenticator.credentialID);
   expect(verification.authenticationInfo?.origin).toEqual(assertionOrigin);
+  expect(verification.authenticationInfo?.rpID).toEqual('dev.dontneeda.pw');
 });
 
 test('should throw when response challenge is not expected value', async () => {
@@ -226,6 +227,7 @@ test('should support multiple possible origins', async () => {
 
   expect(verification.verified).toEqual(true);
   expect(verification.authenticationInfo?.origin).toEqual(assertionOrigin);
+  expect(verification.authenticationInfo?.rpID).toEqual('dev.dontneeda.pw');
 });
 
 test('should throw an error if origin not in list of expected origins', async () => {
@@ -375,6 +377,7 @@ test('should return credential backup info', async () => {
   expect(verification.authenticationInfo?.credentialDeviceType).toEqual('singleDevice');
   expect(verification.authenticationInfo?.credentialBackedUp).toEqual(false);
   expect(verification.authenticationInfo?.origin).toEqual(assertionOrigin);
+  expect(verification.authenticationInfo?.rpID).toEqual('dev.dontneeda.pw');
 });
 
 /**
--- a/packages/server/src/authentication/verifyAuthenticationResponse.test.ts
+++ b/packages/server/src/authentication/verifyAuthenticationResponse.test.ts
@@ -227,7 +227,6 @@ test('should support multiple possible origins', async () => {
 
   expect(verification.verified).toEqual(true);
   expect(verification.authenticationInfo?.origin).toEqual(assertionOrigin);
-  expect(verification.authenticationInfo?.rpID).toEqual('dev.dontneeda.pw');
 });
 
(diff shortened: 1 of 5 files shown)


Mikescops:
@MasterKale thanks for the tip, good to go now 👍

packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -247,6 +251,8 @@ export type VerifiedAuthenticationResponse = {
     userVerified: boolean;
     credentialDeviceType: CredentialDeviceType;
     credentialBackedUp: boolean;
+    origin: string;
+    rpId: string;
```
MasterKale, line 255: I'm sure it's driven people nuts that I do this, but in this project the "D" in the camel-cased "RP ID" is always capitalized:

```suggestion
    rpID: string;
```

packages/server/src/registration/verifyRegistrationResponse.ts
```diff
@@ -289,6 +295,8 @@ export type VerifiedRegistrationResponse = {
     userVerified: boolean;
     credentialDeviceType: CredentialDeviceType;
     credentialBackedUp: boolean;
+    origin: string;
+    rpId?: string;
```
MasterKale, line 299: Same thing here, for consistency's sake:

```suggestion
    rpID?: string;
```

MasterKale (changes requested):

packages/server/src/authentication/verifyAuthenticationResponse.ts
Mikescops: your code is very heterogeneous about this naming, i have checked other interfaces and you use more rpId than rpID

MasterKale: ...Shoot, really? Am I at least consistent in the public API of the core methods? 😅 

I'll take a look when I'm on a computer, I thought I was better about using "`rpID`"...

Mikescops: well I see:

- expectedRPID
- rpIdHash
- matchExpectedRPID
- rpID

but in typescript-types it's `PublicKeyCredentialRequestOptionsJSON.rpId`

let's rename then

MasterKale: In the **public** API for @simplewebauthn/server methods I think I've been consistent with casing on inputs so far:

- `generateRegistrationOptions()`: `rpID`
- `verifyRegistrationResponse()`: `expectedRPID`

- `generateAuthenticationOptions()`: `rpID`
- `verifyAuthenticationResponse()`: `expectedRPID`

The point I was attempting to make is that I think it would be weird to have "`rpID`" as an input, and then get "`rpId"` out. But you've already agreed with me and renamed to `rpID` so I suppose this'll serve as something I can point to in the future whenever someone attempts to introduce `rpId` into the library's public API 😛 

> but in typescript-types it's `PublicKeyCredentialRequestOptionsJSON.rpId`

That's why I supposed that I've probably driven people nuts by going with `rpID` in this package while in the WebAuthn spec they use `rpId`. `PublicKeyCredentialRequestOptionsJSON.rpId` maps to the data structure in the L3 Draft of the spec:

https://w3c.github.io/webauthn/#dictdef-publickeycredentialrequestoptionsjson

In typical use of SimpleWebAuthn, pairing **@simplewebauthn/browser** with **@simplewebauthn/server**, the nice thing is that an RP doesn't need to think about things like this!

packages/server/src/authentication/verifyAuthenticationResponse.test.ts
```diff
@@ -224,6 +226,8 @@ test('should support multiple possible origins', async () => {
   });
 
   expect(verification.verified).toEqual(true);
+  expect(verification.authenticationInfo?.origin).toEqual(assertionOrigin);
+  expect(verification.authenticationInfo?.rpID).toEqual('dev.dontneeda.pw');
```
MasterKale, line 230: We don't need to include this `rpID` expectation in this test since this test is specifically about the `origin` that's returned.

Instead I think it would be better for this expectation to go in the 'should support multiple possible RP IDs'` test below.

packages/server/src/authentication/verifyAuthenticationResponse.test.ts
```diff
@@ -372,6 +376,8 @@ test('should return credential backup info', async () => {
 
   expect(verification.authenticationInfo?.credentialDeviceType).toEqual('singleDevice');
   expect(verification.authenticationInfo?.credentialBackedUp).toEqual(false);
+  expect(verification.authenticationInfo?.origin).toEqual(assertionOrigin);
+  expect(verification.authenticationInfo?.rpID).toEqual('dev.dontneeda.pw');
```
MasterKale, line 380: `origin` and `rpID` aren't really relevant in this test since the test ensures we return credential backup info. The test would be fine without these.

packages/server/src/registration/verifyRegistrationResponse.test.ts
```diff
@@ -496,6 +507,21 @@ test('should support multiple possible origins', async () => {
...
+test('should not set RPID in registrationInfo when not expected', async () => {
+  const verification = await verifyRegistrationResponse({
+    response: attestationNone,
+    expectedChallenge: attestationNoneChallenge,
+    expectedOrigin: 'https://dev.dontneeda.pw',
+    expectedRPID: undefined,
+  });
+
+  expect(verification.verified).toBe(true);
+  expect(verification.registrationInfo?.origin).toEqual('https://dev.dontneeda.pw');
```
MasterKale, line 523: I'd leave this `origin` expectation out of this test because the test specifically says it's testing that an `undefined` expected RP ID will return undefined from the method. The origin isn't here nor there (and has sufficient coverage with all of the other tests you've added in here) that we should be fine to remove this from here.

MasterKale (changes requested):
Thanks for handling all my comments @Mikescops! Just a couple more around refining tests and then I think this is good to go 🚀

packages/server/src/registration/verifyRegistrationResponse.ts
```diff
@@ -289,6 +295,8 @@ export type VerifiedRegistrationResponse = {
     userVerified: boolean;
     credentialDeviceType: CredentialDeviceType;
     credentialBackedUp: boolean;
+    origin: string;
+    rpID?: string;
```
MasterKale, line 299: Note to future-self: I wish this wasn't optional, but alas...it still drives me nuts that FIDO Conformance made me make `expectedRPID` optional in #25.

Mikescops:
I think it's mergeable now 👍 cc @MasterKale

PR #432, opened 2023-08-28
Allow expectedChallenge to return a Promise

### LINKED ISSUE #431: Allow expectedChallenge to return a promise
## Describe the issue

`verifyAuthenticationResponse` and `verifyRegistrationResponse` allow users to provide a function as the value of the `expectedChallenge` property of their `options` parameter. However, this function does not allow for returning a promise, even though both `verifyAuthenticationResponse` and `verifyRegistrationResponse` are async.

If these functions allowed `expectedChallenge` to return a promise, users could perform async validation of `challenges` (e.g. the challenge is asynchronously retrieved from the database, or the challenge included a signature that could be verified by `crypto.subtle.verify` in Deno). Currenlty, only sync validation is possible.

I recommend changing the signature for `expectedChallenge` in `VerifyAuthenticationResponseOpts` and `VerifyRegistrationResponseOpts` to be:
```ts
string | ((challenge: string) => boolean | Promise<boolean>)
```

Then `verifyAuthenticationResponse` and `verifyRegistrationResponse` can perform `await expectedChallenge(challenge)` if it's a function.

jordanbtucker:
This PR allows `verifyAuthenticationResponse` and `verifyRegistrationResponse` to accept `expectedChallenge` properties of their `options` parameter to be Promise returning functions. This allows code like this:

```ts
async function customAsyncVerification(challenge: string): Promise<boolean> {
   ...
}

const isVerified = await verifyAuthenticationResponse({
  ...
  expectedChallenge: customAsyncVerification,
  ...
});
```

Three tests per verification function have been added:
1. Checks for Promises that resolve with `true`.
2. Checks for Promises that resolve with `false`.
3. Checks for Promises that reject.

Fixes #431

### DIFF
--- a/packages/server/src/authentication/verifyAuthenticationResponse.test.ts
+++ b/packages/server/src/authentication/verifyAuthenticationResponse.test.ts
@@ -394,6 +394,82 @@ Deno.test('should fail verification if custom challenge verifier returns false',
   );
 });
 
+Deno.test('should pass verification if custom challenge verifier returns a Promise that resolves with true', async () => {
+  const verification = await verifyAuthenticationResponse({
+    response: {
+      id:
+        'AaIBxnYfL2pDWJmIii6CYgHBruhVvFGHheWamphVioG_TnEXxKA9MW4FWnJh21zsbmRpRJso9i2JmAtWOtXfVd4oXTgYVusXwhWWsA',
+      rawId:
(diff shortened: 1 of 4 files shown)


MasterKale (approved):
Hello @jordanbtucker, thank you for the contribution. This looks good to me.

Issue #438, opened 2023-09-17
Got errors from `startRegistration` with 1Password


unix:
## Describe the issue

With the latest version of 1Password, we are able to register and auth on passkey-enabled websites, [here](https://passkeys-demo.appspot.com/) is an online example from google that works well with 1Password. ([source code](https://github.com/GoogleChromeLabs/passkeys-demo/blob/main/public/client.js))

Icreated a simple example using _SimpleWebAuthn_, but ran into a different error:

### HTTP

We can eventually complete startRegistration using a browser or other device, but it will throw an error (from 1Password).

https://github.com/MasterKale/SimpleWebAuthn/assets/11304944/18f19209-531b-4ed9-b853-0acb5c8b44c1

### HTTPS

With HTTPS, we get an `Illegal invocation` error and the process is blocked.

https://github.com/MasterKale/SimpleWebAuthn/assets/11304944/3d5b1544-ec43-4abf-9615-6dcbc3564bbd

## Reproduction Steps

I've provided an online sample project that replicates this issue: [SimpleWebAuthn-1PD](https://github.com/unix/SimpleWebAuthn-1PD).

1. Download [1Password](https://1password.com/downloads/mac/) & [1Password browser extension](https://1password.com/downloads/browser-extension/);
2. Clone example: `git clone https://github.com/unix/SimpleWebAuthn-1PD`;
3. Install deps: `pnpm install`;
4. Run server: `pnpm serve`;
5. Run web: `pnpm dev`.

We can switch HTTP or HTTPS in the `vite.config.ts` file to view the issue.

## Expected behavior

- It should work as well as this example: [passkeys-demo.appspot.com](https://passkeys-demo.appspot.com/). This is an example of native JavaScript implementation, which can run well with 1PD.

## Code Samples + WebAuthn Options and Responses

- [SimpleWebAuthn-1PD](https://github.com/unix/SimpleWebAuthn-1PD)

## Dependencies

- Browser: chrome 116.0.5845 (Official Build) arm64

### SimpleWebAuthn Libraries

```sh
├── @simplewebauthn/browser@8.0.2 -> ./node_modules/.pnpm/@simplewebauthn+browser@8.0.2/node_modules/@simplewebauthn/browser
├── @simplewebauthn/server@8.1.1 -> ./node_modules/.pnpm/@simplewebauthn+server@8.1.1/node_modules/@simplewebauthn/server
```

## Additional context

[...]

PR #439, opened 2023-09-18
fix: hide errors when get-public-key is not supported

unix:
## Changes

Methods `getPublicKey` and `getPublicKeyAlgorithm` may still return `function type` in unsupported environments and only throw an error when called, for this reason we manually ignore this error.

- Passed the test in the local sample project;
- No side effects seen in the complete registration, auth, etc. process;
- Tested to be able to support 1Password and save data in 1Password (not other devices).

### Refs

- #438

### DIFF
--- a/packages/browser/src/methods/startRegistration.ts
+++ b/packages/browser/src/methods/startRegistration.ts
@@ -67,23 +67,29 @@ export async function startRegistration(
   // L3 says this is required, but browser and webview support are still not guaranteed.
   let responsePublicKeyAlgorithm: number | undefined = undefined;
   if (typeof response.getPublicKeyAlgorithm === 'function') {
-    responsePublicKeyAlgorithm = response.getPublicKeyAlgorithm();
+    try {
+      responsePublicKeyAlgorithm = response.getPublicKeyAlgorithm();
+    } catch (err) {}
   }
-
+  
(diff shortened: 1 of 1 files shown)


MasterKale:
Thank you for the PR, @unix. Can you please add some tests? If you rebase off of master you shouldn't have any issues running **server**'s tests locally now.

zdods (commented):
Excited for this to get in! Ran into this same issue and was thinking it was something with the setup I'm doing - glad to see it'll be remedied.

unix:
@MasterKale 
I rebase to the latest code (from `master`), but I can hardly run all the test cases, which contain a lot of type errors. It seems that the coupling between the organization of the project and the monorepo is a little strange.

I am not sure if the `try catch` will affect the original test cases (it may not seem like it), but I am unable to run existing test cases correctly, making it even more difficult to add test cases. It would be very helpful if you could take over this PR.

aemr3:
@MasterKale we have this issue too. could you please look into this PR? thanks

MasterKale:
I'm going to find time this week to take this work over.

MasterKale:
Thank you again for your contribution @unix. I gave you a shout-out in #443 as it's pretty much this PR but with tests, so I'm marking this one as superseded by the new PR. Thanks for understanding ✌️

PR #455, opened 2023-10-06
Fix `@simplewebauthn/server` not working in edge environments

Maronato:
One of @simplewebauthn/server dependencies cbor-x, [supports streams](https://github.com/kriszyp/cbor-x#streams), which uses Node's and Deno's [`stream`](https://nodejs.org/api/stream.html) interface.

Unfortunately, `stream` is not part of the standard Web APIs that many edge compute runtimes use (e.g. Vercel [doesn't support it](https://nextjs.org/docs/pages/api-reference/edge), and Cloudflare requires a [separate `node:`](https://blog.cloudflare.com/workers-node-js-apis-stream-path/#stream) namespace). Importing it throws errors like these:
Screenshot 2023-10-06 at 13 09 30

Since the only object used from cbor-x is `Encoder`, the fix is pretty simple. Instead of importing the entire  package, we only import its encode file, which is officially exported. This avoids loading the `stream.js` file that uses `stream`.

I tested it by patching @simplewebauthn/server directly in my node_modules and it worked without issues, but I haven't attempted to build it locally. I don't see a reason for it not building, though.

### DIFF
--- a/packages/server/build_npm.ts
+++ b/packages/server/build_npm.ts
@@ -70,8 +70,9 @@ await build({
       name: '@hexagon/base64',
       version: '^1.1.27',
     },
-    'https://deno.land/x/cbor@v1.5.2/index.js': {
+    'https://deno.land/x/cbor@v1.5.2/encode.js': {
       name: 'cbor-x',
+      subPath: 'encode',
       version: '^1.5.2',
     },
     'https://esm.sh/cross-fetch@4.0.0': {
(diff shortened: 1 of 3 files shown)


MasterKale:
Let's see what CI has to say about this, we'll find out quick if this is works. 🤞

MasterKale:
CI failure:

> error: Uncaught "The following specifiers were indicated to be mapped to a package, but were not found:\n  * https://deno.land/x/cbor@v1.5.2/index.js"

You also need to update the path mapping this to an NPM package here: 

https://github.com/MasterKale/SimpleWebAuthn/blob/master/packages/server/build_npm.ts#L73

But since we're not importing **index.js** I don't know hot dnt will handle this 🤔

Maronato:
Looks like we can use [`subPath`](https://github.com/denoland/dnt#mapping-specifier-to-npm-package-subpath). I'm trying to build it now

Maronato:
All tests passed with `act`
Screenshot 2023-10-06 at 14 26 54

Maronato:
Hi @MasterKale!
CI should pass now. Let me know if you need anything else :)

MasterKale:
Thank you for your contribution, @Maronato, this looks good to me :shipit:

MasterKale:
@Maronato This fix is now available in **@simplewebauthn/server@8.3.2**

PR #457, opened 2023-10-13
Fallback webcrypto to global

sergiocarneiro:
Fallback to `globalThis.crypto` when `node:crypto` exists but `crypto.webcrypto` does not.

It also adds the ability to throw `MissingWebCrypto` when `webcrypto` is missing.

This change makes the `getWebCrypto` function more robust and able to work in more runtime setups (e.g. Cloudflare Workers with `node_compat` enabled).

### DIFF
--- a/packages/server/src/helpers/iso/isoCrypto/getWebCrypto.ts
+++ b/packages/server/src/helpers/iso/isoCrypto/getWebCrypto.ts
@@ -11,29 +11,31 @@ export async function getWebCrypto(): Promise<Crypto> {
     return webCrypto;
   }
 
+  /**
+   * Naively attempt to access Crypto as a global object, which popular alternative run-times
+   * support.
+   */
+  // @ts-ignore: ...right here.
+  let _crypto: Crypto = globalThis.crypto;
+
   try {
     /**
      * Naively attempt a Node import...
      */
     // @ts-ignore: We'll handle any errors...
     // dnt-shim-ignore
-    const _crypto = await import('node:crypto');
-    webCrypto = _crypto.webcrypto as Crypto;
-  } catch (_err) {
-    /**
-     * Naively attempt to access Crypto as a global object, which popular alternative run-times
-     * support.
-     */
-    // @ts-ignore: ...right here.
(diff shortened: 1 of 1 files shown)


MasterKale:
> This change makes the `getWebCrypto` function more robust and able to work in more runtime setups (e.g. Cloudflare Workers with `node_compat` enabled).

Hello @sergiocarneiro, I'm wondering what the exact problem is you were having with the current way of picking which `WebCrypto` to use. Are you trying to make `globalThis.crypto` the default the library tries to reach for Node's?

sergiocarneiro:
> I'm wondering what the exact problem is you were having with the current way of picking which `WebCrypto` to use.

Hello @MasterKale. In my Cloudflare Worker, I have Node polyfills enabled. I'm not sure if it's a bug from their platform or not, but I know that what ends up happening is an environment that:
- has `globalThis.crypto`
- allows imports of `node:crypto`
- `crypto.webcrypto` is undefined

So without this change I submitted, using this library (e.g. `getRegistrationOptions`) throws the error `"Cannot read properties of undefined (reading 'getRandomValues')"`.

> Are you trying to make `globalThis.crypto` the default?
Yes, and only override it to `crypto.webcrypto` is it's defined. (maybe I missed a comment there??)

CameronWhiteside:
+1 to this issue--would love to use this in a Cloudflare worker, but without @sergiocarneiro's updates, I'm also fully blocked with the same `"Cannot read properties of undefined (reading 'getRandomValues')"` issue.

MasterKale:
@sergiocarneiro Thank you for the contribution. I took a stab at solving this problem with #468, and I also managed to add test coverage around `getWebCrypto()` that should ensure support for runtimes like your CF worker w/Node compat. In particular this test should capture the nuance of the runtime issue you've been experiencing:

https://github.com/MasterKale/SimpleWebAuthn/pull/468/files#diff-6ac413926c71b2fbaa8af5cffcb57c492adf245dc0d0ec40ec6ba179c537debcR88-R119

Tests just passed over there so I'm going to merge and cut a release with this fix tomorrow.

PR #468, opened 2023-10-27
fix/better-get-web-crypto

MasterKale:
This PR refines the logic inside **server**'s `getWebCrypto()` to support more runtime environments. In particular it should now support CloudFlare Workers with Node polyfills enabled but still missing `import('node:crypto').webcrypto` (this is similar to Node v14 environments, but the CF docs don't explicitly state a target Node version and say they _do_ support `webcrypto` in https://developers.cloudflare.com/workers/runtime-apis/nodejs/crypto/#webcrypto 🤷‍♂️)

Supercedes #457 (this PR also adds test coverage over this trickier method to write tests for.)


MasterKale:
@sergiocarneiro This improved logic in `getWebCrypto()` is now available in **@simplewebauthn/server@8.3.4**. Give it a try and let me know if it solves the issue ✌️

MasterKale:
cc @CameronWhiteside

sergiocarneiro:
@MasterKale It's working ✅

PR #518, opened 2024-02-08
fix(server): ignore `import("node:crypto")` by webpack

### LINKED ISSUE #517: ignore `node:crypto` import with webpack in `@simplewebauthn/server`
Hi, I'm the creator/maintainer of [Auth.js](https://github.com/nextauthjs/next-auth), a runtime/framework agnostic set of auth libraries for modern applications.

We recently added experimental WebAuthn support via your amazing libraries (thank you!), but we realized that our Next.js users would see a `node:crypto` import error when `@simplewebauthn/server` ended up being imported in the Edge runtime: https://github.com/nextauthjs/next-auth/pull/9919

This happened because of the import on this line: https://github.com/MasterKale/SimpleWebAuthn/blob/ba7048cfd791cd9400871d008d21d1354543e2b3/packages/server/src/helpers/iso/isoCrypto/getWebCrypto.ts#L52

Related issues I could find: https://github.com/MasterKale/SimpleWebAuthn/issues/386, https://github.com/MasterKale/SimpleWebAuthn/issues/427, https://github.com/MasterKale/SimpleWebAuthn/issues/471

All modern runtimes have `globalThis.crypto` available, the only exception is Node.js 18. I would like to propose dropping anything related to polyfilling `globalThis.crypto`.

Possible solutions:

1. Drop Node.js 18 support (Support [ends in 1 year](https://endoflife.date/nodejs), arguably would be too big of a change yet, leave it up to you to decide)
2. users on Node.js 18 could be suggested to run Node.js with `--experimental-global-webcrypto`,
3. users on Node.js 18 could be suggested to add `globalThis.crypto ??= await import("node:crypto").webcrypto` in an initialization phase of their code.
4. If the user uses this library via a framework, they might already not have to worry about this. In our case for example, Next.js polyfills (with method 3.) `globalThis.crypto` for us.

**NOTE:** I do understand that this might only be a problem for us currently, but in my opinion, a library should solely rely on its runtime to do polyfilling, if necessary. You are also currently shipping this extra code for all users who are not on Node.js 18. I'm curious if you agree with me! :pray:

If there are more questions or you need more context, let me know!

Thanks for the great work! :green_heart:

balazsorban44:
Fixes #517

### DIFF
--- a/packages/server/src/helpers/iso/isoCrypto/getWebCrypto.ts
+++ b/packages/server/src/helpers/iso/isoCrypto/getWebCrypto.ts
@@ -49,7 +49,7 @@ export const _getWebCryptoInternals = {
   stubThisImportNodeCrypto: async () => {
     try {
       // dnt-shim-ignore
-      const _nodeCrypto = await import('node:crypto');
+      const _nodeCrypto = await import(/* webpackIgnore: true */ 'node:crypto');
       return _nodeCrypto;
     } catch (_err) {
       /**
--- a/packages/server/src/helpers/iso/isoCrypto/getWebCrypto.ts
+++ b/packages/server/src/helpers/iso/isoCrypto/getWebCrypto.ts
@@ -49,6 +49,10 @@ export const _getWebCryptoInternals = {
   stubThisImportNodeCrypto: async () => {
     try {
       // dnt-shim-ignore
+      /**
+       * The `webpackIgnore` here is to help support Next.js' Edge runtime.
+       * See https://github.com/MasterKale/SimpleWebAuthn/issues/517 for more info.
+       */
       const _nodeCrypto = await import(/* webpackIgnore: true */ 'node:crypto');
       return _nodeCrypto;
     } catch (_err) {


MasterKale:
This fix to support Next.js's Edge runtime was identified here:

https://github.com/MasterKale/SimpleWebAuthn/issues/517#issuecomment-1933460685

packages/server/src/helpers/iso/isoCrypto/getWebCrypto.ts
```diff
@@ -49,7 +49,7 @@ export const _getWebCryptoInternals = {
   stubThisImportNodeCrypto: async () => {
     try {
       // dnt-shim-ignore
-      const _nodeCrypto = await import('node:crypto');
+      const _nodeCrypto = await import(/* webpackIgnore: true */ 'node:crypto');
```
MasterKale, line 52: I planned on adding a comment like this before you beat me to the PR for #517:

```suggestion
      /**
       * The `webpackIgnore` here is to help support Next.js' Edge runtime.
       * See https://github.com/MasterKale/SimpleWebAuthn/issues/517 for more info.
       */
      const _nodeCrypto = await import(/* webpackIgnore: true */ 'node:crypto');
```

Once this is in then this is good to go.

MasterKale (approved):
This is good to go, I'll merge this and cut a release tonight 🚀

Issue #513, opened 2024-01-27
Cannot find module 'cbor-x/index-no-eval' or its corresponding type declarations.


mat813:
## Describe the issue

With 9.0.1, I get:

```
$ tsc
node_modules/@simplewebauthn/server/script/deps.d.ts(2,24): error TS2307: Cannot find module 'cbor-x/index-no-eval' or its corresponding type declarations.
```

## Reproduction Steps

1. Update to 9.0.1 on the server
2. Try to build
3. Get the error

## Expected behavior

Not get a build error ?

## Dependencies

- **OS:** Linux
- **Browser:** Node 21.6
- **Authenticator:** n/a

### SimpleWebAuthn Libraries

Server and types at 9.0.1

MasterKale:
Hello @mat813, thanks for the report. This is related to #511, weird that an issue like this wasn't caught locally or in CI. I'll dig into this and report back.

Can you share your project's **tsconfig.json**?

sualko:
I have the same issue.

```
{
  "ts-node": {
    "files": true
  },
  "compilerOptions": {
    "module": "commonjs",
    "strict": true,
    "strictPropertyInitialization": false,
    "noImplicitOverride": true,
    "declaration": true,
    "removeComments": true,
    "resolveJsonModule": true,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "esModuleInterop": true,
    "allowSyntheticDefaultImports": true,
    "target": "es6",
    "lib": ["dom", "dom.iterable", "esnext"],
    "sourceMap": true,
    "outDir": "./dist",
    "baseUrl": "./",
    "incremental": true,
    "paths": {
      "@/*": ["./src/*"]
    }
  },
  "exclude": ["node_modules", "dist", "tmp"],
  "include": ["src", "test"]
}
```

Not sure if this is related, but for some reason the default export no longer works for me. Instead of

```
import SimpleWebAuthnServer from '@simplewebauthn/server';
```

I have to use

```
import * as SimpleWebAuthnServer from '@simplewebauthn/server';
```

We are running Typescript in version 5.3.3.

sualko:
It seams like the package is not in the npm registry, but I'm not entirely sure what is going on here:

```
$ npm i cbor-x/index-no-eval
npm ERR! code 128
npm ERR! An unknown git error occurred
npm ERR! command git --no-replace-objects ls-remote ssh://git@github.com/cbor-x/index-no-eval.git
[...]

PR #521, opened 2024-02-15
fix/513-cbor-x-type-issues

MasterKale:
This PR replaces use of the **cbor-x** dependency with **tiny-cbor**. The use of `cbor-x/index-no-eval` caused issues with transpiling typical Node-centric use of SimpleWebAuthn because TypeScript couldn't find types for `index-no-eval` due to the project's architecture.

No reasonable workaround within SimpleWebAuthn could be determined so I have decided to move to the excellent **tiny-cbor** instead. This library supports the right amount of CBOR to work with WebAuthn responses, and also cross-publishes on https://deno.land/x and NPM, so it'll be a great replacement.

Addresses #513.


LeviSchuck:
Looks correct to me

Issue #458, opened 2023-10-13
`verifyAuthenticationResponse` TypeError: cosePublicKey.get is not a function


H3RSKO:
## Describe the issue

When trying to verify an authenticator with `verifyAuthenticationResponse`, i'm getting `TypeError: cosePublicKey.get is not a function`.

Stack trace:
<img width="1011" alt="Screenshot 2023-10-13 at 4 49 03 PM" src="https://github.com/MasterKale/SimpleWebAuthn/assets/34112131/ddb470bb-9fee-4334-baaa-dade1bbe26e4">

## Reproduction Steps

The registration flow works as expected, but I cannot complete the final `verifyAuthenticationResponse` step.  

This is the data i am passing into `verifyAuthenticationResponse`:

```
{
  response: {
    id: 'zoJI0Aa2dOb2a4qx3Kzm0IjGgxl8ZBnXf1Ta6Om4gEI',
    rawId: 'zoJI0Aa2dOb2a4qx3Kzm0IjGgxl8ZBnXf1Ta6Om4gEI',
    response: {
      authenticatorData: 'YkwHDjAspXeq2sn_IIWVunr0oFvhhtyCJubZrA56lk0FAAAAAA',
      clientDataJSON: 'eyJ0eXBlIjoid2ViYXV0aG4uZ2V0IiwiY2hhbGxlbmdlIjoiX3pmTTRLZ2s1WEk3Y1EyQmo4RURaQWtHQnZ0d2RqQnpadm5TZmVNSHlTWSIsIm9yaWdpbiI6Imh0dHBzOi8vbXlsb2NhbHdlYnNpdGUiLCJjcm9zc09yaWdpbiI6ZmFsc2V9',
      signature: 'MEUCIFAsp5nX_KwwguG6LI5PisY1R4rYPi0eAPiqjRstGj6cAiEA7Au2txVKIP58A1RkFHvNEthhM22tmO7JaYfVoogF2TQ',
      userHandle: '7a8da293-bc06-43a8-b7ab-0a27f14e3c82'
    },
    type: 'public-key',
    clientExtensionResults: {},
    authenticatorAttachment: 'platform'
  },
  expectedChallenge: '_zfM4Kgk5XI7cQ2Bj8EDZAkGBvtwdjBzZvnSfeMHySY',
  expectedOrigin: 'https://mylocalwebsite',
  expectedRPID: 'mylocalwebsite',
  requireUserVerification: true,
  authenticator: {
[...]

Issue #345, opened 2023-02-04
WebAuthn works locally, but not in production


joshtblack:
Hi there, I've setup WebAuthn and got it working great locally, so in preparation for pushing to prod, I changed the rpID and origin respectively. This works semi-okay, on production, registration of both passkeys (Macbook Touch ID in this instance) and security keys work. However, when I go to authenticate with a security key, I receive this error:

`NotSupportedError: Unrecognized name.`

from verifyAuthenticationResponse. If I attempt to login with a passkey, it works fine.

This is a non-issue on local, so I'm hoping you'd be able to shed some light onto what this error means?

Thank you!

MasterKale:
I'll need to see some code. For now please show me how you're calling `generateAuthenticationOptions()` or `startAuthentication()` (it's unclear if you're using SimpleWebAuthn end-to-end, or only using one or the other.) Right now it sounds to me like this is an issue within the browser, before `verifyAuthenticationResponse()` ever gets called, so it's probably something about the options that are making it to the actual WebAuthn API call within `startAuthentication()`.

joshtblack:
Sure. Here's how I'm calling generateAuthenticationOptions():

`const user = await collection.findOne({id: req.session!.getUserId()});
    const rpID =
        process.env.NODE_ENV === "development"
            ? "localhost"
            : "dashboard.file.glass";
    const userAuthenticators = user?.u2f?.authenticators;

    const options = generateAuthenticationOptions({
        timeout: 60000,
        userVerification: 'preferred',
        rpID,
    });

const response = {
        ...options,
        allowCredentials: userAuthenticators.map((authenticator: { credentialID: any, transports: any }) => ({
            id: authenticator.credentialID,
            type: 'public-key'
        })),
    }
    res.status(200).json(response);`

You might be wondering why I'm adding the 'allowCredentials' in the response rather than the options itself - it's because for some reason if I add the exact same allowCredentials code in the options, the id value is always empty, adding it into the response itself is the only way I could get it to work. Maybe you could shed some light on that as well?

Here's how I am calling startAuthentication():
`asseResp = await startAuthentication(response.data);
  const verificationResponse = await axios.post("/api/user/account/u2f/authentication/verifyAuthenticationResponse", asseResp);`

joshtblack:
I'll include the code snippet to startAuthentication as well:

```
const response = await axios.get("/api/user/account/u2f/authentication/generateAuthenticationOptions");
        let asseResp;
[...]

PR #524, opened 2024-02-21
Fixed generateAuthenticationOptions

H3RSKO:
As discussed [here](https://github.com/MasterKale/SimpleWebAuthn/issues/458) and [here](https://github.com/MasterKale/SimpleWebAuthn/issues/345#issuecomment-1585735973), even though we are storing `credentialID` as a `base64Url`, the `generateAuthenticationOptions` function requires it to be a buffer, and _then_ converts it back to a `base64Url`. 

This PR adds the ability to `generateAuthenticationOptions` to check if it is a string (i.e base64Url) and if it is will return it as is, if it is not a string (i.e it is a buffer) it will go through the existing conversion of buffer to base64Url. This allows users who are storing credentialID as a base64Url (which is recommended), to pass it in directly to generateAuthenticationOptions without having to convert it to a buffer first.

### DIFF
--- a/packages/server/src/authentication/generateAuthenticationOptions.ts
+++ b/packages/server/src/authentication/generateAuthenticationOptions.ts
@@ -53,7 +53,7 @@ export async function generateAuthenticationOptions(
     challenge: isoBase64URL.fromBuffer(_challenge),
     allowCredentials: allowCredentials?.map((cred) => ({
       ...cred,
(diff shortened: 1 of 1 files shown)


MasterKale:
Thank you for the contribution @H3RSKO, but I'm closing this because I'll be solving this problem more comprehensively as part of addressing #528 (which was inspired by your PR.)

PR #526, opened 2024-02-21
chore(server,helpers): change `decodeClientDataJSON` signature to accept `Base64URLString`

### LINKED ISSUE #525: Change signature of `decodeClientDataJSON` to accept `Base64URLString` as paramater
## Describe the issue

At the moment, `decodeClientDataJSON` accepts a `string` as a parameter. In order to communicate effectively what kind of string it expects to receive, we should change the type to `Base64URLString`, as that's the type of `clientDataJSON` in `AuthenticatorAssertionResponseJSON`.

```ts
const credential = await startAuthentication(optionsJson);

if (!credential) {
  return null;
}

const decodedClientData = decodeClientDataJSON(credential.response.clientDataJSON) // `clientDataJSON` is of type Base64URLString`
```

carlos-menezes:
Closes #525


MasterKale:
Superseded by #527.

PR #533, opened 2024-02-26
Update getWebCrypto for Node 20 (Approach 1)

### LINKED ISSUE #532: Update `getWebCrypto()` for Node 20
## Describe the issue

Once #519 raises the minimum supported Node version to v20 then `getWebCrypto()` can consistently access `WebCrypto` at `globalThis.crypto` instead of having to `require('node:crypto')`. This is an issue to track this work while I continue working on the v10 release.

Node docs related to using `globalThis.crypto` to access `WebCrypto`: 

https://nodejs.org/docs/latest-v20.x/api/webcrypto.html#web-crypto-api

MasterKale:
This PR updates `getWebCrypto()` to access `globalThis.crypto` exclusively since Node 20 supports this as a stable API access points. No attempts are made to import from `node:crypto` anymore.

This is a **bigger** PR because `getWebCrypto()` becomes synchronous, which spiders out into a few files that use `WebCrypto`. I hope I don't regret this...

Fixes #532.

### DIFF
--- a/packages/server/src/authentication/generateAuthenticationOptions.test.ts
+++ b/packages/server/src/authentication/generateAuthenticationOptions.test.ts
@@ -1,6 +1,6 @@
 import { assert, assertEquals, assertExists } from 'https://deno.land/std@0.198.0/assert/mod.ts';
 
-import { isoBase64URL, isoUint8Array } from '../helpers/iso/index.ts';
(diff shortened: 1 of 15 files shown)


MasterKale:
Superseded by #535

PR #559, opened 2024-04-14
Update README.md

outbackStack:


MasterKale:
Thank you for the submission @outbackStack, but this project isn't open to outside contributions at the moment even for small stuff like this. I've gone ahead and committed a similar change in the README to help sync things up 🙇

PR #594, opened 2024-07-19
Fix ECDSA Signature Unwrapping

nlordell:
We ran into some unexpected and inconsistent signature verification errors. After some digging I traced it down to what I believe is an issue in the `unwrapEC2Signature` method.

From what I understood, this is responsible for taking the ASN.1 encoded ECDSA signature, and encoding it into a format that the WebCrypto library expects (which is `r ‖ s` - where `‖` is the byte concatenation operator). In particular the Web Crypto API documents[^1] that:

> 4. Let `n` be the smallest integer such that `n * 8` is greater than the logarithm to base 2 of the order of the base point of the elliptic curve identified by `params`.
> 5. Convert `r` to an octet string of length `n` and append this sequence of bytes to `result`.
> 6. Convert `s` to an octet string of length `n` and append this sequence of bytes to `result`.

So, this means that the signature components have very specific expected byte lengths when "unwrapping" the signature for use with the Web Crypto "subtle" API. In practice, I found that it was not always producing signatures of the correct length, and in particular the `rBytes` and `sBytes` values were not always of length `n` as defined above. Specifically, I was working with P-256 signatures and noticed that sometimes `r` and/or `s` had a length of 31 instead of 32 (which is the smallest byte-length required to represent an integer equal to the order of the P-256 curve).

Looking into ASN.1 specification[^2]:

> **8.3.2** If the contents octets of an integer value encoding consist of more than one octet, then the bits of the first octet and bit 8 of the second octet:
> 1. shall not all be ones; and
> 2. shall not all be zero.
>
> _NOTE – These rules ensure that an integer value is always encoded in the smallest possible number of octets._

After more closely inspecting the signature I was using, it would occasionally produce signatures where the MSB of the `r` or `s` value was `0`, which would cause the BER encoding of the signature component to be shorter than expected. From the P-256 test case, the signature in hex can be broken down to:

```
# SEQUENCE(03) of length 0x44(68)
30 44
# INTEGER(03) of length 0x1f(31)
02 1f 587f17b4cf4cc561fe4c1bfecb230f4ff653bedb6d76b3cf6bb1f8fea484b4
# INTEGER(03) of length 0x31(33)
02 21 00aeb47b3d78206408830fdf203d9a63512847c577e4449a1198d866c7a33682ef
```

Note that, the leading `00` byte from the signature `r` value is removed in the ASN.1 encoding, while an additional leading `00` byte is added to the `s` value (so that it isn't confused with `02 20 aeb47b3d78206408830fdf203d9a63512847c577e4449a1198d866c7a33682ef`, which would represent `-0x514b84c287df9bf77cf020dfc2659caed7b83a881bbb65ee672799385cc97d11`).

This PR fixes the `unwrapEC2Signature` implementation in order such that it computes `r ‖ s` where the byte length of `r` and `s` are guaranteed to be the same (and specifically `n` as it is defined in the Web Crypto API specification[^1]).

I added some test vectors to make sure things work (that would fail on `master`). I also added a P-521 test vector which does not work (as it does not seem to be implemented by Deno). I left it in as `ignored` as the vector itself should be correct (as manually porting the test to run in my browser passed) and it might be nice to add once P-521 is supported.

[^1]: <https://www.w3.org/TR/WebCryptoAPI/#ecdsa-operations>
[^2]: <https://www.itu.int/rec/T-REC-X.690-202102-I/en>

### DIFF
--- a/packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
+++ b/packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
@@ -56,7 +56,7 @@ function toNormalizedBytes(i: ArrayBuffer, n: number): Uint8Array {
   const normalizedBytes = new Uint8Array(n);
   if (iBytes.length <= n) {
     normalizedBytes.set(iBytes, n - iBytes.length);
-  } else if (iBytes.length === n + 1 && iBytes[0] === 0) {
+  } else if (iBytes.length === n + 1 && iBytes[0] === 0 && (iBytes[1] & 0x80) === 0x80) {
     normalizedBytes.set(iBytes.slice(1));
   } else {
     throw new Error("invalid signature component length");
--- a/packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
+++ b/packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
@@ -9,19 +9,25 @@ import { isoUint8Array } from '../index.ts';
  */
 export function unwrapEC2Signature(signature: Uint8Array, crv: COSECRV): Uint8Array {
   const parsedSignature = AsnParser.parse(signature, ECDSASigValue);
-  const n = getSignatureComponentLength(crv);
+  const rBytes = new Uint8Array(parsedSignature.r);
+  const sBytes = new Uint8Array(parsedSignature.s);
 
-  const rBytes = toNormalizedBytes(parsedSignature.r, n);
-  const sBytes = toNormalizedBytes(parsedSignature.s, n);
+  const componentLength = getSignatureComponentLength(crv);
+  const rNormalizedBytes = toNormalizedBytes(rBytes, componentLength);
+  const sNormalizedBytes = toNormalizedBytes(sBytes, componentLength);
 
-  const finalSignature = isoUint8Array.concat([rBytes, sBytes]);
+  const finalSignature = isoUint8Array.concat([
+    rNormalizedBytes,
+    sNormalizedBytes,
+  ]);
 
   return finalSignature;
 }
 
 /**
- * ECDSA signatures with in the subtle crypto API expect signatures with `r` and `s` values
- * encoded to a specific length depending on the order of the curve.
+ * The SubtleCrypto Web Crypto API expects ECDSA signatures with `r` and `s` values to be encoded
+ * to a specific length depending on the order of the curve. This function returns the expected
+ * byte-length for each of the `r` and `s` signature components.
  *
  * See <https://www.w3.org/TR/WebCryptoAPI/#ecdsa-operations>
  */
@@ -42,24 +48,28 @@ function getSignatureComponentLength(crv: COSECRV): number {
  * Converts the ASN.1 integer representation to bytes of a specific length `n`.
  *
  * DER encodes integers as big-endian byte arrays, with as small as possible representation and
- * require leading `0` bytes to disambiguate between negative and positive numbers. This means
- * that `r` and `s` can potentially not be the expected length `n` that is needed by the WebCrypto
- * subtle API: if there it leading `0`s it can be shorter than expected, and if it has a leading
- * `1` bit, it can be one byte longer.
+ * requires a leading `0` byte to disambiguate between negative and positive numbers. This means
+ * that `r` and `s` can potentially not be the expected byte-length that is needed by the
+ * SubtleCrypto Web Crypto API: if there are leading `0`s it can be shorter than expected, and if
+ * it has a leading `1` bit, it can be one byte longer.
  *
  * See <https://www.itu.int/rec/T-REC-X.690-202102-I/en>
  * See <https://www.w3.org/TR/WebCryptoAPI/#ecdsa-operations>
  */
-function toNormalizedBytes(i: ArrayBuffer, n: number): Uint8Array {
-  const iBytes = new Uint8Array(i);
-
-  const normalizedBytes = new Uint8Array(n);
-  if (iBytes.length <= n) {
-    normalizedBytes.set(iBytes, n - iBytes.length);
-  } else if (iBytes.length === n + 1 && iBytes[0] === 0 && (iBytes[1] & 0x80) === 0x80) {
-    normalizedBytes.set(iBytes.slice(1));
+function toNormalizedBytes(bytes: Uint8Array, componentLength: number): Uint8Array {
+  let normalizedBytes;
+  if (bytes.length < componentLength) {
+    // In case the bytes are shorter than expected, we need to pad it with leading `0`s.
+    normalizedBytes = new Uint8Array(componentLength);
+    normalizedBytes.set(bytes, componentLength - bytes.length);
+  } else if (bytes.length === componentLength) {
+    normalizedBytes = bytes;
+  } else if (bytes.length === componentLength + 1 && bytes[0] === 0 && (bytes[1] & 0x80) === 0x80) {
+    // The bytes contain a leading `0` to encode that the integer is positive. This leading `0`
+    // needs to be removed for compatibility with the SubtleCrypto Web Crypto API.
+    normalizedBytes = bytes.subarray(1);
   } else {
-    throw new Error("invalid signature component length");
+    throw new Error(`invalid signature component length ${bytes.length} (expected ${componentLength})`);
   }
 
   return normalizedBytes;
--- a/packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
+++ b/packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
@@ -1,4 +1,5 @@
 import { AsnParser, ECDSASigValue } from '../../../deps.ts';
+import { COSECRV } from '../../cose.ts';
 import { isoUint8Array } from '../index.ts';
 
 /**
@@ -6,18 +7,12 @@ import { isoUint8Array } from '../index.ts';
  *
  * See https://www.w3.org/TR/webauthn-2/#sctn-signature-attestation-types
  */
-export function unwrapEC2Signature(signature: Uint8Array): Uint8Array {
+export function unwrapEC2Signature(signature: Uint8Array, crv: COSECRV): Uint8Array {
   const parsedSignature = AsnParser.parse(signature, ECDSASigValue);
-  let rBytes = new Uint8Array(parsedSignature.r);
-  let sBytes = new Uint8Array(parsedSignature.s);
+  const n = getSignatureComponentLength(crv);
 
-  if (shouldRemoveLeadingZero(rBytes)) {
-    rBytes = rBytes.slice(1);
-  }
-
-  if (shouldRemoveLeadingZero(sBytes)) {
-    sBytes = sBytes.slice(1);
-  }
+  const rBytes = toNormalizedBytes(parsedSignature.r, n);
+  const sBytes = toNormalizedBytes(parsedSignature.s, n);
 
   const finalSignature = isoUint8Array.concat([rBytes, sBytes]);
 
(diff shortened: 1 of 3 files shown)


MasterKale:
Hello @nlordell, thank you very much for the submission. My past-self has required me to start off by saying that this project is not open to external contributions as per the README.

_That said_, given that I don't work with WebCrypto day-to-day, the code in the PR looks high quality to me, and your rationale is explained quite well, I'm going to make an exception. Crypto is hard to do right and it seems to be dumb luck that I've managed to get away this long with EC2 signature wrapping as it's been.

Thank you again for bringing this to my attention, and for helping improve use of WebCrypto in SimpleWebAuthn 🙇 

Tests are passing so I'm going to start reviewing the PR "this weekend" (I might try for tonight, otherwise sometime before Monday.) After a quick perusal just now I have only a few nitpicks here and there, otherwise the bulk of the code looks good to me.

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
```diff
@@ -1,36 +1,66 @@
...
+ * DER encodes integers as big-endian byte arrays, with as small as possible representation and
+ * require leading `0` bytes to disambiguate between negative and positive numbers. This means
+ * that `r` and `s` can potentially not be the expected length `n` that is needed by the WebCrypto
+ * subtle API: if there it leading `0`s it can be shorter than expected, and if it has a leading
+ * `1` bit, it can be one byte longer.
+ *
+ * See <https://www.itu.int/rec/T-REC-X.690-202102-I/en>
+ * See <https://www.w3.org/TR/WebCryptoAPI/#ecdsa-operations>
+ */
+function toNormalizedBytes(i: ArrayBuffer, n: number): Uint8Array {
```
MasterKale, line 53: I'd like to see these parameter names to be a bit more descriptive:

```suggestion
function toNormalizedBytes(bytes: ArrayBuffer, componentLength: number): Uint8Array {
```

MasterKale, line 53: Does it negatively impact the functionality of this method if this `ArrayBuffer` type gets swapped for a `Uint8Array` type? I'd prefer to keep things consistent and use `Uint8Array` bytes everywhere (at least in **@simplewebauthn/server**), and since `i` gets fed immediately into `Uint8Array` why not keep the argument the same type since we're making a copy of the bytes anyway?

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
```diff
@@ -1,36 +1,66 @@
...
+  const iBytes = new Uint8Array(i);
+
+  const normalizedBytes = new Uint8Array(n);
+  if (iBytes.length <= n) {
+    normalizedBytes.set(iBytes, n - iBytes.length);
```
MasterKale, line 58: So if the bytes are Big Endian, this operation here could lead to many leading zeroes before the contents of `iBytes`:

```js
const length = 6;
const iBytes = new Uint8Array([1,2,3]);

const normalized = new Uint8Array(length);

normalized.set(iBytes, length - iBytes.length);

// Uint8Array(6) [ 0, 0, 0, 1, 2, 3 ]
console.log(normalized);
```

I'm mainly confirming my understanding of what's going on here.

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
```diff
@@ -1,36 +1,66 @@
...
+  } else if (iBytes.length === n + 1 && iBytes[0] === 0 && (iBytes[1] & 0x80) === 0x80) {
+    normalizedBytes.set(iBytes.slice(1));
```
MasterKale, line 60: I think it's possible to reduce the number of copies of the input bytes that are made if there's more consistent use of `.slice()` (I'm assuming the method arguments are renamed as I requested above):

```suggestion
  const normalizedBytes = new Uint8Array(n);
  
  if (bytes.length <= componentLength) {
    normalizedBytes.set(bytes.slice(0), componentLength - bytes.length);
  } else if (bytes.length === componentLength + 1 && bytes[0] === 0 && (bytes[1] & 0x80) === 0x80) {
    normalizedBytes.set(bytes.slice(1));
```

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
```diff
@@ -1,36 +1,66 @@
...
+  } else {
+    throw new Error("invalid signature component length");
```
MasterKale, line 62:
```suggestion
    throw new Error(`invalid signature length: ${bytes.length} bytes (component length: ${componentLength}`);
```

nlordell:
> My past-self has required me to start off by saying that this project is not open to external contributions as per the README.

Oh shoot, I'm really sorry, should have checked before making a PR 🙈. With this in mind, if you feel strongly about sticking to your rule, feel free to close this PR and implement the change as you see fit (and borrow some of the code from the PR if you like) - in a way consider this PR as a detailed issue report 😅. I definitely won't be disappointed and am happy in being able to contribute a detailed bug finding 😄.

> dumb luck that I've managed to get away this long with EC2 signature wrapping as it's been.

Just some wild speculation, but maybe ECDSA credentials aren't as widely used as EdDSA ones? Also, it only happens when the leading 8 bits are 0, so a roughly 1/256 chance (well, a little less because `r` and `s` aren't in the full range of a 256-bit integer), so it is fairly intermittent.

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
nlordell: I would still say "invalid signature **component** length" since this is the length of the `r` or `s` value and not the whole signature.

Adding more debugging info makes sense!

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
```diff
@@ -1,36 +1,66 @@
...
  * In WebAuthn, EC2 signatures are wrapped in ASN.1 structure so we need to peel r and s apart.
  *
  * See https://www.w3.org/TR/webauthn-2/#sctn-signature-attestation-types
  */
-export function unwrapEC2Signature(signature: Uint8Array): Uint8Array {
+export function unwrapEC2Signature(signature: Uint8Array, crv: COSECRV): Uint8Array {
   const parsedSignature = AsnParser.parse(signature, ECDSASigValue);
-  let rBytes = new Uint8Array(parsedSignature.r);
-  let sBytes = new Uint8Array(parsedSignature.s);
+  const n = getSignatureComponentLength(crv);
```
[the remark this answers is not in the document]
nlordell: Sounds good. My original rationale was to name it the same as in the Web Crypto API specification, but your suggested naming is also nice.

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
nlordell: This is my understanding as well.

Given the example above, we would have `r` be computed as the value `0x000000010203` and `r` generally needs 6 bytes in order to fully represent all its possible values. However, because of ASN.1 BER encoding rules, this should get written as `02 /*integer*/ 03 /*length*/ 010203 /*value*/`, and so here, we need to pad the bytes with leading 0s so that its representation is the expected length `6` expected by the Web Crypto API.

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
nlordell: > Does it negatively impact the functionality of this method if this `ArrayBuffer` type gets swapped for a `Uint8Array` type?

I don't think so. I remember that when I checked, the type of both `parsedSignature.r` and `.s` are `ArrayBuffer`, and hence the type of this parameter. We can also change it so that we wrap `parsedSignature.r` and `.s` into `Uint8Array`s before passing it into this function if you prefer.

nlordell:
Thanks for the thoughtful review! I will squash the suggestions into a single commit and play around with your idea of reducing input byte copies when I'm back home tonight.

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
MasterKale: Ah, `r` and `s` are called "components", one day I'll finish this [Real-World Cryptography book](https://www.manning.com/books/real-world-cryptography) I swear 🫠 

I'll defer to you on exact verbiage here, my intent here was to add lengths to the error message to help library users better troubleshoot this error 😅

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
MasterKale: > We can also change it so that we wrap `parsedSignature.r` and `.s` into `Uint8Array`s before passing it into this function if you prefer.

Looking at the code that's being replaced here, I'd like to go with this suggestion (since the new code is doing the same thing with iBytes)

Let's bring these lines back after the ASN parsing:

image

And then change the `ArrayBuffer` type to `Uint8Array` on the method argument.

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
nlordell: > Ah, r and s are called "components"

Just for transparency, this isn't a technical term I found in literature, just a more general term that fits here... I tried to find what people actually call `r` and `s` individually, but couldn't find any real consensus.

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
nlordell: I implemented something I'm happy with here, where the underlying buffer only gets copied if the signature component needs to be extended. In the more common cases with the signature components are already the correct length (or have a leading `0`), we reuse the underlying buffer without copying.

Note that instead of using `slice` here, I opted for `subarray` so that the underlying buffer does not get copied.

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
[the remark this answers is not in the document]
nlordell: Implemented in 83d716162ed95bee92dd787df52b5f61477354cc

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
nlordell: Implemented in 83d716162ed95bee92dd787df52b5f61477354cc

nlordell: Implemented in 83d716162ed95bee92dd787df52b5f61477354cc

packages/server/src/helpers/iso/isoCrypto/unwrapEC2Signature.ts
nlordell: I added a bit more text to the comments so the meaning of "component" _hopefully_ clearer

nlordell:
I implemented the suggestions from the PR review - quick question, do you have any preference for formatting for the file? I figured you were using `deno fmt`, but this changes the quoting style to use double quotes which is different from the neighboring source files.

Let me know if you want me to run an automated code formatter on this.

MasterKale:
> I implemented the suggestions from the PR review - quick question, do you have any preference for formatting for the file? I figured you were using `deno fmt`, but this changes the quoting style to use double quotes which is different from the neighboring source files.
> 
> Let me know if you want me to run an automated code formatter on this.

Looking at my command history, in the past I've run this from the root of the monorepo:

```sh
$> deno fmt packages/server/src/
```

Husky git pre-commit hooks are _supposed_ to take care of this for you but I'm not surprised they didn't work for you - I haven't had great confidence in them in the past and it seems not much has improved about them.

nlordell:
Thanks for the tip! Changes should be formatted now.

MasterKale (approved):

MasterKale:
@nlordell This all looks good to me, thank you for lending this project some of your SubtleCrypto expertise 🙇

MasterKale:
@nlordell I've released **@simplewebauthn/server@10.0.1** that has this fix in it. Thank you again! 🚀

Issue #582, opened 2024-06-13
Add support for WebAuthn's conditional create


MasterKale:
## Describe the issue

At WWDC 24 this week Apple debuted **macOS Sequioa Dev Beta 1**, which includes **Safari 18.0 Beta**. According to the WebKit blog, Safari 18.0 Beta supports "conditional create" added to WebAuthn L3. This is what Apple is referring to when it mentions "automatic passkey upgrades."

This new capability allows an RP to "silently" trigger a passkey registration after a successful traditional authentication (i.e. username+password auth) without the user needing to click through any browser modals. From that point on the RP could switch the user to passkeys at the next login.

Supporting this isn't much different than supporting conditional UI during auth; it's an otherwise normal call to `navigator.credentials.create()` with some specific options passed in. I can easily add such a capability to **@simplewebauthn/browser**'s `startRegistration()` method similar to how I added conditional UI support with `startAuthentication(..., true)`.

## Additional Resources

Safari flow chart for when an RP can trigger conditional create:

![Screenshot 2024-06-11 at 1 17 58 PM](https://github.com/MasterKale/SimpleWebAuthn/assets/5166470/9c56c31d-fec1-4ea2-a2c0-0d6e114eb08c)

- https://webkit.org/blog/15443/news-from-wwdc24-webkit-in-safari-18-beta/
- https://w3c.github.io/webauthn/#sctn-createCredential
- https://developer.apple.com/wwdc24/10125?time=38

lmarschall:
Hey @MasterKale,

a few months ago, I was trying to implement the conditional UI flow with the browser package, but I wasn't able to get it working properly without eventual fails.

https://simplewebauthn.dev/docs/packages/browser#browser-autofill-aka-conditional-ui

As you described here, we should add the `startAuthentication` workflow in the head section or let the UI wait some n seconds, is this workaround still required, or are there any better solutions, because I find it quite unpleasant to implement the `startAuthentication` in the head section or wait n seconds :disappointed:

lmarschall:
@MasterKale 

Would it be possible to create a new passkey on login with an existing passkey as well?

MasterKale:
> As you described here, we should add the `startAuthentication` workflow in the head section or let the UI wait some n seconds, is this workaround still required, or are there any better solutions, because I find it quite unpleasant to implement the `startAuthentication` in the head section or wait n seconds 😞

Good question @lmarschall, let me ask around and see what the consensus is.

[...]

PR #621, opened 2024-10-06
fix/597-shadow-dom-support

### LINKED ISSUE #597: The check for browser auto fill doesn't take into account shadow DOMs (ie: web components)
The check here doesn't dive into the shadow DOM of web components so it won't see if there's an autocomplete input in a web component. 

https://github.com/MasterKale/SimpleWebAuthn/blob/a4a178cc8daa4eb5106b9708456195f3aff9643f/packages/browser/src/methods/startAuthentication.ts#L59

While this is supported in the major browsers now: https://issues.chromium.org/issues/40251770

MasterKale:
This PR adds a new `verifyBrowserAutofillInput` options to @simplewebauthn/browser's `startAuthentication()`. When set to `false` the method will no longer raise an error when it cannot find a suitable `<input autocomplete="... webauthn">` element.

Most users of this method shouldn't need to use this new flag. However projects using e.g. shadow components which contain an otherwise suitable element for conditional UI can use this new flag to benefit from `startAuthentication()` as well.

Fixes #597.

### `startAuthentication()` now uses an option object

Positional arguments have been replaced by a single object containing all options. Keywords match the name of the previously positional arguments.

To update existing implementations, wrap existing options in an object with corresponding properties:

**Before:**

```ts
startAuthentication(options, true);
```

**After:**

```ts
startAuthentication({ optionsJSON: options, useBrowserAutofill: true });
```

### `startRegistration()` now uses an option object

Positional arguments have been replaced by a single object containing all options. Keywords match the name of the previously positional arguments.

To update existing implementations, wrap existing options in an object with corresponding properties:

**Before:**

```ts
startRegistration(options);
```

**After:**

```ts
startRegistration({ optionsJSON: options });
```

(It may seem premature to make this change to `startRegistration()` too, but this is in preparation for #582)

### DIFF
--- a/packages/browser/src/methods/startAuthentication.test.ts
+++ b/packages/browser/src/methods/startAuthentication.test.ts
@@ -72,7 +72,7 @@ afterEach(() => {
 });
 
 test('should convert options before passing to navigator.credentials.get(...)', async () => {
-  await startAuthentication(goodOpts1);
+  await startAuthentication({ optionsJSON: goodOpts1 });
 
   const argsPublicKey = mockNavigatorGet.mock.calls[0][0].publicKey;
   const credId = argsPublicKey.allowCredentials[0].id;
@@ -87,8 +87,10 @@ test('should convert options before passing to navigator.credentials.get(...)', 
 
 test('should support optional allowCredential', async () => {
   await startAuthentication({
-    challenge: '1T6uHri4OAQ',
-    timeout: 1,
+    optionsJSON: {
+      challenge: '1T6uHri4OAQ',
+      timeout: 1,
+    },
   });
 
   expect(mockNavigatorGet.mock.calls[0][0].allowCredentials).toEqual(undefined);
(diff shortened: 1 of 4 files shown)


treeder:
Curious why you didn't call it `options` instead of `optionsJSON`? It's not actually a JSON string is it?

MasterKale:
> Curious why you didn't call it `options` instead of `optionsJSON`? It's not actually a JSON string is it?

Historically that option has been called `optionsJSON` to try and communicate that the input is **not** standard WebAuthn options (e.g. `PublicKeyCredentialCreationOptions`) but the JSON-ified output of them from the corresponding `generate...Options()` method in **@simplewebauthn/server** (e.g. `PublicKeyCredentialCreationOptionsJSON` that just arrived in the browser as JSON.)

I debated a different name but figured the goal remains unchanged so I'm sticking with it for now.

PR #638, opened 2024-11-12
Example - FIDO Fix

dmaesj:
ENABLE_CONFORMANCE="true" is currently broken on the example project. This is due to an outdated model in the fido-conformance.ts file. Updated the user model to be the same as used in index.ts


MasterKale:
@dmaesj Thank you for the flagging this problem, and for going the extra step and creating a PR. [As per the README](https://github.com/MasterKale/SimpleWebAuthn?tab=readme-ov-file#contributions), though this project is not currently open to outside contributions. Please create an issue instead and I'll be sure to fix this.

PR #713, opened 2025-08-18
Improve TypeScript type narrowing

SantaClaas:
Hi, thanks for providing this library. I ran across a TypeScript issue with the type `VerifiedRegistrationResponse` where Typescript complains `registrationInfo` might be undefined even though `verified` is `true`.
However from looking at the code it seems like `registrationInfo` can never be undefined when `verified` is `true`.

This PR changes the Types to reflect that behavior and adjusts the return of `verifyRegistrationResponse` to fix resulting TypeScript errors where TypeScript can't ensure the types otherwise.

No behavior is changed to my knowledge and I kept the changes to a minimum. Though maybe it makes sense to split out a type for `registrationInfo`. 
It might create new errors and thus break existing type checks. But I see those as a feature as they might uncover bugs. 😅

### DIFF
--- a/packages/server/src/registration/verifyRegistrationResponse.ts
+++ b/packages/server/src/registration/verifyRegistrationResponse.ts
@@ -275,36 +275,39 @@ export async function verifyRegistrationResponse(
     throw new Error(`Unsupported Attestation Format: ${fmt}`);
   }
 
-  const toReturn: VerifiedRegistrationResponse = {
-    verified,
-  };
-
-  if (toReturn.verified) {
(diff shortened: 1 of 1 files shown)


MasterKale:
Hello @SantaClaas, thank you for going through the trouble of making this PR. I agree that this is a DX pain point with a simple solution. I'll go ahead and make this change.

[As per the README](https://github.com/MasterKale/SimpleWebAuthn?tab=readme-ov-file#contributions) this project is not currently accepting PRs from external contributors. I more than welcome anyone to create **issues** for things like this, though, and in this case I think this is such a non-controversial improvement that I went ahead and created #714 to track this issue.

PR #717, opened 2025-08-22
Upgrade project to Deno v2.4

MasterKale:
This PR refactors **@simplewebauthn/browser** and **@simplewebauthn/server** to support use in Deno projects running Deno v2.2+.

Deno v2.2 launched with TypeScript 5.7. In this version of TypeScript, `Uint8Array` was made generic, to be of type `Uint8Array<ArrayBufferLike>`. Existing values typed simply as `Uint8Array` now needed to be refactored to specify the type of their array buffer:

**Before**
```ts
const value: Uint8Array = new Uint8Array();
```

**After**
```ts
const value: Uint8Array<ArrayBuffer> = new Uint8Array<ArrayBuffer>();
```

Unfortunately "simply refactoring all instances of Uint8Array" to specify a buffer type would break the code in Deno v2.1 and earlier where Uint8Array is not generic.

To support both runtimes, I took a page out of Deno's handbook and defined a new `Uint8Array_` type instead. This type is equivalent to non-generic `Uint8Array` in Deno v2.1 and earlier, and equivalent to `Uint8Array<ArrayBuffer>` in Deno v2.2 and beyond.

I'm not entirely happy with how this bit of Deno runtime changes gets exposed in the public API of the project. I'm keeping my fingers crossed that it'll be transparent to most everyone. A more succinct copy of all of the context above has been included in the docstring for `Uint8Array_` for those who go digging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ request new features, or to suggest changes to existing features.
 
 Install the following before proceeding:
 
-- **Deno v2.1.x**
+- **Deno v2.4.x**
 
 After pulling down the code, set up dependencies:
 


P4sca1:
Hey @MasterKale, for my implementation this change introduced a type error.
`credential.publicKey` now expects `Uint8Array<ArrayBuffer>`, but my database driver (prisma) returns `Uint8Array<ArrayBufferLike>`. For now I resolved the type error, by instantiating a new Uint8Array with the same content:

```ts
// Before:
publicKey:passkey.publicKey,

// After:
publicKey: new Uint8Array(passkey.publicKey),
```

PR #759, opened 2026-05-06
Support cross-origin authentication verification

agektmr:
To support cross-origin authentication on SimpleWebAuthn, this pull request adds the following to the `verifyAuthenticationResponse.ts`:
* Add a new `expectedTopOrigin` property to the `options`
* Check if `crossOrigin` is `true` or `false`
* Check if `topOrigin` meets the expectation

### DIFF
--- a/packages/server/src/authentication/verifyAuthenticationResponse.test.ts
+++ b/packages/server/src/authentication/verifyAuthenticationResponse.test.ts
@@ -603,7 +603,7 @@ Deno.test('should throw when crossOrigin is true, topOrigin is missing, and expe
           credential,
         }),
       Error,
-      'Unexpected cross-origin request',
+      'Unexpected cross-origin authentication response',
     );
   } finally {
     mockDecodeClientData.restore();
@@ -766,7 +766,7 @@ Deno.test('should throw when crossOrigin is true and topOrigin does not match an
           credential,
         }),
       Error,
-      'Unexpected cross-origin authentication within "https://wrong.top.origin.com", expected one of: https://top.origin.com, https://other.origin.com',
+      'Unexpected cross-origin authentication response origin "https://wrong.top.origin.com", expected one of: https://top.origin.com, https://other.origin.com',
     );
   } finally {
     mockDecodeClientData.restore();
@@ -806,7 +806,7 @@ Deno.test('should throw when crossOrigin is true but expectedTopOrigin is not sp
   }
 });
 
-Deno.test('should NOT check topOrigin when crossOrigin is false', async () => {
+Deno.test('should throw when topOrigin is set despite crossOrigin being false', async () => {
   const mockDecodeClientData = stub(
     _decodeClientDataJSONInternals,
     'stubThis',
@@ -822,16 +822,19 @@ Deno.test('should NOT check topOrigin when crossOrigin is false', async () => {
   );
 
   try {
-    const verification = await verifyAuthenticationResponse({
-      response: assertionResponse,
-      expectedChallenge: assertionChallenge,
-      expectedOrigin: assertionOrigin,
-      expectedRPID: 'dev.dontneeda.pw',
-      credential,
-      requireUserVerification: false,
-    });
-
-    assertEquals(verification.verified, true);
+    await assertRejects(
+      () =>
+        verifyAuthenticationResponse({
+          response: assertionResponse,
+          expectedChallenge: assertionChallenge,
+          expectedOrigin: assertionOrigin,
+          expectedRPID: 'dev.dontneeda.pw',
+          credential,
+          requireUserVerification: false,
+        }),
+      Error,
+      'Unexpected top origin without cross origin request',
+    );
   } finally {
     mockDecodeClientData.restore();
   }
--- a/packages/server/src/authentication/verifyAuthenticationResponse.test.ts
+++ b/packages/server/src/authentication/verifyAuthenticationResponse.test.ts
@@ -578,6 +578,232 @@ Deno.test('should return user verified flag after successful auth', async () => 
   assertFalse(verification.authenticationInfo?.userVerified);
 });
 
+// Deno.test('should throw when crossOrigin is true but topOrigin is missing', async () => {
+//   const mockDecodeClientData = stub(
+//     _decodeClientDataJSONInternals,
+//     'stubThis',
+//     returnsNext([
+//       {
+//         type: 'webauthn.get',
+//         origin: assertionOrigin,
+//         challenge: assertionChallenge,
+//         crossOrigin: true,
+//       },
+//     ]),
+//   );
+// 
+//   await assertRejects(
+//     () =>
+//       verifyAuthenticationResponse({
+//         response: assertionResponse,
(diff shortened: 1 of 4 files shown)


packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -120,6 +122,32 @@ export async function verifyAuthenticationResponse(
...
 
+  // Check that the authentication response is within an expected iframe
+  if (crossOrigin) {
+    // TODO: Since Safari doesn't support `topOrigin` as of May 2026, only check this when `topOrigin` is available for now.
+    if (topOrigin) {
+      if (Array.isArray(expectedTopOrigin)) {
+        if (!expectedTopOrigin.includes(topOrigin)) {
+          const joinedExpectedTopOrigin = expectedTopOrigin.join(', ');
+          throw new Error(
+            `Unexpected cross-origin authentication within "${topOrigin}", expected one of: ${joinedExpectedTopOrigin}`,
```
MasterKale: A bit of a nitpick:

```suggestion
            `Unexpected cross-origin authentication response origin "${topOrigin}", expected one of: ${joinedExpectedTopOrigin}`,
```

packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -120,6 +122,32 @@ export async function verifyAuthenticationResponse(
...
+        if (topOrigin !== expectedTopOrigin) {
+          throw new Error(
+            `Unexpected cross-origin authentication within "${topOrigin}", expected: ${expectedTopOrigin}`,
+          );
+        }
+      }
+    } else if (!expectedTopOrigin) {
+      // If `expectedTopOrigin` is not set, this is an unexpected cross-origin request.
+      throw new Error(
+        'Unexpected cross-origin request',
```
MasterKale: Trying to stay consistent on "authentication response" in error messages:

```suggestion
        'Unexpected cross-origin authentication response',
```

packages/server/src/authentication/verifyAuthenticationResponse.test.ts
```diff
@@ -578,6 +578,265 @@ Deno.test('should return user verified flag after successful auth', async () =>
...
+        }),
+      Error,
+      'Unexpected cross-origin authentication within "https://top.origin.com", expected: undefined',
+    );
+  } finally {
+    mockDecodeClientData.restore();
+  }
+});
+
+Deno.test('should NOT check topOrigin when crossOrigin is false', async () => {
```
MasterKale: This test is an interesting one. According to the spec, `topOrigin` should only be set if `crossOrigin` is `true`:

> It is set only if the call was made from context that is not [same-origin with its ancestors](https://w3c.github.io/webappsec-credential-management/#same-origin-with-its-ancestors), i.e. if [crossOrigin](https://w3c.github.io/webauthn/#dom-collectedclientdata-crossorigin) is true.

https://w3c.github.io/webauthn/#dom-collectedclientdata-toporigin

Based on this I'd actually make the new code error out if `crossOrigin` is `false` but there's a `topOrigin` defined in `clientDataJSON` 🤔

packages/server/src/authentication/verifyAuthenticationResponse.ts
agektmr: "response origin" feels odd as it's the iframe origin that's actually responding. How about "top origin"?

agektmr: Actually, this makes sense. Never mind.

packages/server/src/authentication/verifyAuthenticationResponse.test.ts
agektmr: Fixed. Please have a look.

packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -120,6 +122,37 @@ export async function verifyAuthenticationResponse(
...
+      throw new Error(
+        'Unexpected cross-origin authentication response',
+      );
+    }
+  } else if (topOrigin) {
+    // If `topOrigin` is set despite `crossOrigin` being false, this is an unexpected request.
+    throw new Error(
+      'Unexpected top origin without cross origin request',
+    );
+  }
```
MasterKale, line 167: Apologies, the suggestion isn't showing the more nuanced changes I'm proposing to this block of logic. However I sat down and followed the logic flow and came up with a few changes I think will help with the readability of this nuanced chunk of code.

I also sought to leave some notes for ourselves to remember why we're doing it all this way, and also make the errors a bit more verbose in the name of making them actionable for RPs that encounter them 🤔

```suggestion
  // Check that the authentication response is within an expected iframe
  if (crossOrigin) {
    /**
     * TODO: Since Safari doesn't support `topOrigin` as of May 2026, only check this when
     * `topOrigin` is available for now.
     */
    if (topOrigin) {
      if (!expectedTopOrigin) {
        /**
         * If `expectedTopOrigin` is not set, this should be considered an unexpected cross-origin
         * request. Reject the response while helping the RP understand how they need to update this
         * method call if they want to support verification of such WebAuthn authentication requests.
         */
        throw new Error(
          `Detected cross-origin authentication response from top origin of "${topOrigin}", but a value for \`expectedTopOrigin\` was not specified when calling \`verifyAuthenticationResponse()\``,
        );
      }
      
      if (Array.isArray(expectedTopOrigin)) {
        if (!expectedTopOrigin.includes(topOrigin)) {
          const joinedExpectedTopOrigin = expectedTopOrigin.join(', ');
          throw new Error(
            `Unexpected cross-origin authentication response top origin of "${topOrigin}", expected one of: ${joinedExpectedTopOrigin}`,
          );
        }
      } else {
        if (topOrigin !== expectedTopOrigin) {
          throw new Error(
            `Unexpected cross-origin authentication response top origin of "${topOrigin}", expected: ${expectedTopOrigin}`,
          );
        }
      }
    }
  } else {
    if (topOrigin) {
      /**
       * If `topOrigin` is set despite `crossOrigin` being false, this is an unexpected request.
       * 
       * See https://w3c.github.io/webauthn/#dom-collectedclientdata-toporigin.
       */
      throw new Error(`Unexpected top origin of "${topOrigin}" within a non-cross-origin authentication response. This error should be reported to the browser vendor as a WebAuthn specification violation with a link to https://w3c.github.io/webauthn/#dom-collectedclientdata-toporigin`);
    }
  }
```

MasterKale (commented):
@agektmr I'm so sorry for taking so long to respond. I think this PR is really close to getting merged. I had a few proposed changes I'd like you to consider first.

agektmr:
@MasterKale the suggestion looks great. Thank you!

MasterKale (approved):
Sorry for the delay, this looks great. I want to roll this out as a new feautre with the v14.0.0 milestone (https://github.com/MasterKale/SimpleWebAuthn/milestone/54) so please bear with me as I coordinate development of another feature or two before rolling this out 🙇‍♂️

PR #765, opened 2026-05-27
Improve MetadataService logging

### LINKED ISSUE #751: Refactor MetadataService to use an injected logger
## Describe the issue
I haven't been all that satisfied with how MetadataService handles logging, so it's been a no-op for over two years now:

https://github.com/MasterKale/SimpleWebAuthn/blob/78ad52f44afe43647fa9d70fa7fd889e07fd1107/packages/server/src/helpers/logging.ts#L17-L20

I researched alternative solutions and landed on a solution: update `MetadataService.initialize()` to accept a new "`Logger`" object that allows the project using SimpleWebAuthn to capture logs appropriate to its setup, without having to accommodate any library-specific solution.

The interface would look something like this:

```ts
interface Logger {
  debug:   (message: string, ...args: unknown[]) => void;
  info:    (message: string, ...args: unknown[]) => void;
  warning: (message: string, ...args: unknown[]) => void;
  error:   (message: string, ...args: unknown[]) => void;
}
```

MasterKale:
This PR introduces a new `logger` argument to `MetadataService.initialize()`. This argument is of type `SimpleWebAuthnLogger`, which is a simple object of logging methods:

```ts
export interface SimpleWebAuthnLogger {
  debug: (message: string, ...args: unknown[]) => void;
  info: (message: string, ...args: unknown[]) => void;
  warn: (message: string, ...args: unknown[]) => void;
  error: (message: string, ...args: unknown[]) => void;
}
```

The intention is that projects using SimpleWebAuthn can define an instance of this `interface` that logs the message and arbitrary arguments in a way that makes sense for that project. This eliminates any SimpleWebAuthn-specific logging concerns that led to the old implementation getting ripped out of this project.

Right now only `MetadataService` will make use of this new logging pattern.

Fixes #751.

### DIFF
--- a/packages/server/src/helpers/index.ts
+++ b/packages/server/src/helpers/index.ts
@@ -15,3 +15,4 @@ export * from './verifySignature.ts';
 export * from './iso/index.ts';
 export * from '../metadata/verifyMDSBlob.ts';
 export * as cose from './cose.ts';
+export { type SimpleWebAuthnLogger } from './logging.ts';
--- a/packages/server/src/helpers/logging.test.ts
+++ b/packages/server/src/helpers/logging.test.ts
@@ -1,0 +1,23 @@
+import { assertEquals } from '@std/assert';
+import { assertSpyCall, spy } from '@std/testing/mock';
+
+import { buildLoggerAllMethods } from './logging.ts';
+
+Deno.test('should define default methods for undefined methods', () => {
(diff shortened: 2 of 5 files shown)


MasterKale:
Marking WIP for now until I write some tests. I'm otherwise happy with the current implementation.

PR #779, opened 2026-07-09
Support "Immediate UI" for authentication

darthmaim:
This adds support for "Immediate UI" to the `startAuthentication` function by adding a new `useImmediateUIMode` option.

Here are some explainers for Immediate UI (but I'm sure you are already aware 😉):
 - https://developer.chrome.com/docs/identity/immediate-ui-mode?hl=en
 - https://developer.chrome.com/blog/webauthn-immediate-ui?hl=en
 - https://github.com/w3c/webauthn/blob/main/explainers/immediate-mediation.md
 
This might be too soon to add as only Chrome (but already stable) supports this and there is still some movement in the API. It might make more sense to land this once the API settles and might require some breaking changes to the options to allow different combinations of `uiMode` and `mediation`.

Let me know if you are open to add this yet or have any API changes, then I will take the time to add some tests.

---

A different approach might be to allow passing custom options that are spread into `getOptions` (or maybe a callback that gets called with the constructed `getOptions` and returns modified ones) so its the users responsibility to use APIs that are not yet natively supported by `startAuthentication`.

### DIFF
--- a/packages/browser/src/methods/startAuthentication.ts
+++ b/packages/browser/src/methods/startAuthentication.ts
@@ -20,12 +20,14 @@ export type StartAuthenticationOpts = Parameters<typeof startAuthentication>[0];
  * @param optionsJSON Output from **@simplewebauthn/server**'s `generateAuthenticationOptions()`
  * @param useBrowserAutofill (Optional) Initialize conditional UI to enable logging in via browser autofill prompts. Defaults to `false`.
  * @param verifyBrowserAutofillInput (Optional) Ensure a suitable `<input>` element is present when `useBrowserAutofill` is `true`. Defaults to `true`.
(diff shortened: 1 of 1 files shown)


MasterKale:
Thanks for the PR, @darthmaim, but this project is [not currently open to external PRs](https://github.com/MasterKale/SimpleWebAuthn#contributions) as per the README.

I do invite you to open an issue to track immediate UI support, though, as I agree it would be cool to support it in SimpleWebAuthn ✌️

darthmaim:
Ah my bad, I somehow missed that. I've opened #780 to track Immediate UI.

---

Btw, you can disable PRs from outside contributors in GitHub these days:
image

MasterKale:
Thanks for the reminder about disabling PRs, it turns out I'd temporarily re-enabled it and forgot to turn it back off 🤦‍♂️ 

Sorry for the confusion.

PR #800, opened 2026-09-04
Fix X.509 parsing to support PQC use in attestation statements

### LINKED ISSUE #799: Attestations with X.509 certs containing PQC public keys fails to verify
## Describe the issue

This was reported to me via backchannel. It turns out I'd neglected to map X.509 PQC algorithm OIDs, and @peculiar/x509 didn't properly support X.509 certificates with PQC algorithm identifiers itself, so PQC support was only partially achieved in v14.0.0.

I discovered it was because the PQC test vectors I could include in the library were `"none"` type attestation so they didn't test these codepaths. And my manual testing clearly skipped requesting attestation or this would have been caught earlier.

The solution is to pull in the newer @peculiar/x509@2.1.0 library (and a related ASN.1 schema library, @peculiar/asn1-x509-post-quantum) and map the OIDs.

MasterKale:
This PR updates dependencies and maps PQC X.509 algorithm OIDs so that attestation statements leveraging PQC algorithms can be properly verified.

Note that I still cannot yet include full registration responses containing PQC-using attestation statements as test vectors because I am under an NDA.

This PR also includes some intentional raising of a new `PQCNotSupportedError` when verifying responses with PQC passkeys in runtimes that don't support PQC. This happens earlier in the flow so that downstream problems that happen because of the lack of PQC support are't masked as other types of errors.

And I also included some test optimizations.

Fixes #799


packages/server/src/authentication/verifyAuthenticationResponse.ts
```diff
@@ -283,14 +289,27 @@ export async function verifyAuthenticationResponse(
...
+
   const { credentialDeviceType, credentialBackedUp } = parseBackupFlags(flags);
 
   const toReturn: VerifiedAuthenticationResponse = {
-    verified: await verifySignature({
-      signature,
-      data: signatureBase,
-      credentialPublicKey: credential.publicKey,
-    }),
+    verified,
```
MasterKale: I can revert this to slim down the PR

